PR #2, opened 2018-10-10
Update Bintray Package Info

uilianries:
Hi!

This PR is about a plugin to update Bintray package page.

For each new package created, we need to fill manually Bintray page, but all needed data could be extract from the recipe.

Related issue: https://github.com/bincrafters/community/issues/476


danimtb:
I really like the idea of this new plugin as Bintray still does not fill metadata automatically, but we will have to wait to make a decision on how plugins will be managed for installation, versioning and activation and what will be the long-term purpose of repository. Please, keep the PR open and I will try to discuss it with the team. Many thanks!!!

jgsogo:
We have added Travis and AppVeyor CI to this repo and an example test for the `attribute_checker` hook. We need to merge master into this PR (to get CI) and then implement some tests to check that the hook is working (it should be easy to mock `requests.get` function, but ping me if you need help).

Thanks!

uilianries:
@jgsogo the bintray hook is updated!

Since it requires bintray api and interface validation I would prefer to create a fake package to validate all steps:

https://bintray.com/conan-community/test-distribution/foo%3Abar

You will need to set `CONAN_LOGIN_USERNAME_VIRTUAL` and `CONAN_PASSWORD_VIRTUAL`. Both values are the same used for conan community login.

TIL: If you need to get an environment variable when running tox, just use `passenv` instead of spending 2h trying all kind of workaround ...

uilianries:
Ops!

```
ERROR: [HOOK - conan-center.py] post_build(): 'settings.os' doesn't exist
'settings' possible configurations are ['arch_build', 'os_build']
Traceback (most recent call last):
  File "/opt/miniconda3/lib/python3.6/site-packages/conans/client/hook_manager.py", line 56, in execute
    method(output, **kwargs)
  File "/home/uilian/.conan/hooks/conan-center.py", line 121, in post_build
    if not _files_match_settings(conanfile, conanfile.build_folder):
  File "/home/uilian/.conan/hooks/conan-center.py", line 209, in _files_match_settings
    if conanfile.settings.os == "Windows":
  File "/opt/miniconda3/lib/python3.6/site-packages/conans/model/settings.py", line 257, in __getattr__
    self._check_field(field)
  File "/opt/miniconda3/lib/python3.6/site-packages/conans/model/settings.py", line 253, in _check_field
    raise undefined_field(self._name, field, self.fields, self._parent_value)
conans.errors.ConanException: 'settings.os' doesn't exist
'settings' possible configurations are ['arch_build', 'os_build']
```

I need to support installer packages too.

README.md
```diff
@@ -16,6 +16,16 @@ Place your hook Python files under *~/.conan/hooks*. The name of the hook would
 *~/.conan/hook/conan-center.py
 ```
 
+Only copying plugin files will not activate them.
+
+## Conan config as installer
+
+To install all plugins from Conan plugins:
+
+``$ conan config install https://github.com/conan-io/plugins``
```
jgsogo:
```suggestion
``$ conan config install https://github.com/conan-io/hooks``
```

README.md
```diff
@@ -16,6 +16,16 @@ Place your hook Python files under *~/.conan/hooks*. The name of the hook would
 *~/.conan/hook/conan-center.py
 ```
 
+Only copying plugin files will not activate them.
```
jgsogo:
```suggestion
Only copying hook files will not activate them.
```

README.md
```diff
@@ -16,6 +16,16 @@ Place your hook Python files under *~/.conan/hooks*. The name of the hook would
...
+
+## Conan config as installer
+
+To install all plugins from Conan plugins:
```
jgsogo:
```suggestion
To install all hooks from Conan repository in Github:
```

README.md
```diff
@@ -16,6 +16,16 @@ Place your hook Python files under *~/.conan/hooks*. The name of the hook would
...
+
+``$ conan config install https://github.com/conan-io/plugins``
+
+Conan config install does not activate any plugins.
```
jgsogo:
```suggestion
Conan config install does not activate any hook.
```

README.md
```diff
@@ -30,6 +40,11 @@ If you handle multiple dependencies in your project is better to add a *conan.co
     conan-center
 ```
 
+## Supported plugins
```
jgsogo:
```suggestion
## Supported plugins
```
```suggestion
## Supported hooks
```

README.md
```diff
@@ -30,6 +40,11 @@ If you handle multiple dependencies in your project is better to add a *conan.co
...
+
+- [Conan Center](plugins/conan-center.py): Conan recipe linter
```
jgsogo:
```suggestion
- [Conan Center](hooks/conan-center.py): Conan recipe linter
```

hooks/bintray-update.py
```diff
@@ -0,0 +1,180 @@
+# -*- coding: utf-8 -*-
+"""Bintray Package info update
+
+This Conan plugin reads your recipe and update it Bintray package info using the attributes.
```
jgsogo:
```suggestion
This Conan plugin reads your recipe and updates its Bintray package info using the attributes.
```

jgsogo:
Having a look at this PR I have some doubts about the test suite. Do you think it is ok to call Conan CLI or shall we just test the hook function with mocked arguments? 

In the beginning, I thought that it would be ok to use Conan CLI, this way we can test our hooks against different versions of Conan itself (it is what I did with the `attribute_checker` hook), but now I see that this approach has several drawbacks in order to mock functions: as we are running an external tool through command line, this is no longer our program and we cannot mock function calls :/, so:

 1. we can use `conan_api` (undocumented, not stable) to avoid using CLI, or
 1. we can test just the hook function using fake arguments.

Each approach has pros and cons:
 1. `conan_api` is not stable and it might be modified, we can encapsulate these modifications inside a [helper conan client](https://github.com/conan-io/hooks/blob/master/tests/utils/test_cases/conan_client.py) that we will need to maintain and it will get more and more confusing; but we will be sure that all the Conan workflow will be taken into account for our hooks.

    We will need to run all tests against all (supported) versions of Conan, like it is implemented now.

2. Faster, easier and safer, nothing external to the hook itself will affect the tests. But to keep consistency and check that every Conan version will work, we would need some extra tests, calling the Conan CLI to check that the parameters that get to the hooks functions are the expected ones (the ones we will be mocking in the actual tests).

    No need to run the tests related to the hooks against different Conan versions, only the tests which check that the arguments that reach the hooks functions are the expected ones.

In any case, I think that we should mock the calls to external services, we can store sample json responses (also failing ones) and our test suite won't rely on Bintray status. 
What about if Bintray changes its API? Well, I suppose it won't happen too often, but we can run tests against Bintray API and check our stored jsons against the returned ones, this way we will know if a failing test is due to our code or to Bintray changes.

What do you think about this dissertation?

jgsogo:
Talking about second option: testing just the hook function with mocked arguments

 * we can mock `requests.get` function with something like:

   ```python
   from mock import patch
   ...

   def side_effect(url, ...):
      # Read proper file depending on URL passed as argument
      return file_json

   # Prepare the arguments for the hook
   # Mock the requests object
   with patch.object("requests.get", side_effect= side_effect):
      # Run the hook function
      post_upload(arguments,...)
   ```

There are still integration tests:
 * [Required] run a single test for every conan version (like it is implemented now in the CI), using a `test_framework_hook` that checks that the inputs for every hook are the expected
 * [Optional] check that the json returned from Bintray and the one stored in the files that we are using to mock `requests.get` are equal.

uilianries:
@jgsogo I made some changes following your points. Thanks for the example, it helped me to understand how to mock requests :+1: 

Okay, the new test calls the hook directly, and all Bintray's answers are mocked up. I've tried to follow Bintray status code and messages -- All commands were executed to check Bintray's response.

tests/test_hooks/test_update_bintray.py
```diff
@@ -0,0 +1,133 @@
...
+            "maturity": "Stable"
+        }
+        return MockResponse(response, 200, True, text=json.dumps(response))
+
+    return MockResponse(None, 404, text="URL must be declared")
+
+def side_effect_patch(*args, **kwargs):
+    url = kwargs.get("url")
+    auth = kwargs.get("auth")
+    json_data = kwargs.get("json")
```
jgsogo: Why not...?

```python

def side_effect_patch(url, auth, json, *args, **kwargs):
    ...
```

tox.ini
```diff
@@ -11,17 +11,22 @@ deps =
...
     coverage: coverage-enable-subprocess
     pytest>=3.6
+    mock>=2.0.0
+    webtest>=2.0.32
 
 setenv =
     PYTHONDONTWRITEBYTECODE=1
     PYTHONPATH = {toxinidir}:{env:PYTHONPATH:}
+    CONAN_USERNAME = "chespirito"
+    CONAN_PASSWORD = "elchapulincolorado"
```
jgsogo: No need here, I would add those variables to the context of the `post_upload_recipe` call.

tox.ini
```diff
@@ -11,17 +11,22 @@ deps =
     conandev: https://github.com/conan-io/conan/archive/develop.tar.gz
     coverage: coverage-enable-subprocess
     pytest>=3.6
+    mock>=2.0.0
+    webtest>=2.0.32
```
jgsogo: is it used?

tests/test_hooks/test_update_bintray.py
```diff
@@ -0,0 +1,133 @@
...
+import os
+import re
+import textwrap
+import json
+from mock import mock
+
+from conans import tools
+from conans.client.remote_registry import Remote
+from conans.model.ref import ConanFileReference
+from conans.test.utils.tools import TestBufferConanOutput
```
jgsogo: I'm sure that we should only depend on the public interface of Conan, to what it is exposed in the main `conans` module: `tools`, `ConanFile`,... but we need to go inside to get some useful stuff. I'll come up with something in the following days, leave it as it is for now. 👍

tests/test_hooks/test_update_bintray.py
uilianries: done

tox.ini
uilianries: yes, https://github.com/conan-io/hooks/pull/2/files#diff-8eebc7307c1e61c5efaea1449796443cR7

tests/test_hooks/test_update_bintray.py
uilianries: okay

tox.ini
uilianries: done

uilianries:
@jgsogo I've updated the unit tests, but as hooks doesn't have `__init__.py`, I've added hooks dir in the path: https://github.com/conan-io/hooks/pull/2/files#diff-8eebc7307c1e61c5efaea1449796443cR19

hooks/bintray_update.py
```diff
@@ -0,0 +1,179 @@
...
+        remote_info = _get_package_info_from_bintray(package_url=package_url)
+        output.info("Inspecting recipe info ...")
+        recipe_info = _get_package_info_from_recipe(conanfile_path=conanfile_path)
+        updated_info = _update_package_info(recipe_info=recipe_info, remote_info=remote_info)
+        if updated_info:
+            output.info("Bintray is outdated. Updating Bintray package info ...")
+            _patch_bintray_package_info(package_url=package_url, package_info=updated_info, remote=remote)
+        else:
+            output.info("Bintray package info is up-to-date.")
+    except Exception as error:
```
SSE4: isn't it too broad catch?

hooks/bintray_update.py
```diff
@@ -0,0 +1,179 @@
...
+def _extract_user_repo(remote):
+    """
+    Extract username and repository name from remote URL
+    :param remote: Conan remote object
+    :return: username, repository name
+    """
+    pattern = r'https?:\/\/api.bintray.com\/conan\/(.*)\/(.*)'
+    match = re.match(pattern=pattern, string=remote.url)
+    if not match:
+        raise Exception("Could not extract subject and repo from %s" % remote.url)
```
SSE4: something more specific, e.g. `ValueError`?

hooks/bintray_update.py
```diff
@@ -0,0 +1,179 @@
...
+
+def _get_package_info_from_bintray(package_url):
+    """
+    Read package info from Bintray repository
+    :param package_url: Bintray package URL
+    :return: Package data from Bintray
+    """
+    response = requests.get(url=package_url)
+    if not response.ok:
+        raise Exception("Could not request package info: {}".format(response.text))
```
SSE4: more specific (e.g. `HTTPError`)?

hooks/bintray_update.py
```diff
@@ -0,0 +1,179 @@
...
+        updated_info['vcs_url'] = url
+
+    issue_tracker_url = "{}/community/issues".format(url[:url.rfind('/')]) if url else ""
+    issue_tracker_url = os.getenv("BINTRAY_ISSUE_TRACKER_URL", issue_tracker_url)
+
+    if issue_tracker_url:
+        if remote_info['issue_tracker_url'] != issue_tracker_url:
+            updated_info['issue_tracker_url'] = issue_tracker_url
+
+    if 'maturity' not in remote_info or remote_info['maturity'] != "Stable":
```
SSE4: maybe check `channel` here and only set maturity to the stable if channel is stable?

hooks/conan-center.py
```diff
@@ -232,3 +233,9 @@ def _default_package_structure(folder):
         if folder not in default_folders:
             return False
     return True
+
+
+def _get_os(conanfile):
+    if hasattr(conanfile.settings, "os"):
+        return conanfile.settings.os
+    return conanfile.settings.os_build
```
SSE4: this will probably fail if recipe has neither `os` nor `os_build`?

hooks/bintray_update.py
```diff
@@ -0,0 +1,179 @@
...
+        if remote_info['website_url'] != homepage:
+            updated_info['website_url'] = homepage
+
+    return updated_info
+
+
+def _patch_bintray_package_info(package_url, package_info, remote):
+    username, password = _get_credentials(remote)
+    response = requests.patch(
+        url=package_url, json=package_info, auth=HTTPBasicAuth(username, password))
```
SSE4: can you confirm is only `basic` auth supported by bintray so far? no `digest`?

tests/test_hooks/test_update_bintray.py
```diff
@@ -0,0 +1,145 @@
...
+
+from conans import tools
+
+from conans.client.remote_registry import Remote
+from conans.model.ref import ConanFileReference
+from conans.test.utils.tools import TestBufferConanOutput
+
+from tests.utils.test_cases.conan_client import ConanClientTestCase
+
+sys.path.append(os.path.join(os.getenv("TOX_WORK_DIR"), '..', 'hooks'))
```
jgsogo: We should remove this in the future, let's see how this repo layout looks like in the following days.

hooks/bintray_update.py
uilianries: Yes, I'll print which properties will be updated.

hooks/bintray_update.py
uilianries: excellent observation, but how could we know if is it stable branch? CPT offers a mechanism to check it based on branch name, but hooks does not require CPT.

hooks/conan-center.py
uilianries: yeah, but for header-only cases this method will not be executed. Do know any scenario where os is not required but is not header-only too? Maybe we could just return None.

hooks/bintray_update.py
SSE4: probably "stable/xxx" by default and add some env var to override it?

hooks/conan-center.py
SSE4: may be some installer or python requires package?

uilianries: I think gyp_installer is a good example, but it's configured as header-only. Python requires is another good example, so yes, we should return None when both are available or raise an error.

hooks/bintray_update.py
uilianries: According CTP we could use CONAN_STABLE_BRANCH_PATTERN as env var, but we could accept stable and release as well. WDYT? BTW if we not set to Stable maturity, so we could set to Development instead.

SSE4: `CONAN_STABLE_BRANCH_PATTERN` is good one to use, also `stable` and `release` are both good as default pattern.
for the `Development` maturity, do I right understand that maturity applies to the all package versions? at least I see it on bintray GUI as global. in that case upload of testing after stable will override global maturity. so I'd prefer to uploading non-stable as not touching maturity field at all.

uilianries: you are right, maturity is a global property. I'll follow your advice, I think will be better change only when stable.

hooks/bintray_update.py
uilianries: I've confirmed now, only Basic Auth is supported :(
If you read the documentation: https://bintray.com/docs/api/#Authentication, it says:

> The Bintray REST API requires an applicative API key. An API key can be obtained from the user profile page. Authentication is achieved using HTTP Basic Authentication with the user’s name as username and the API key as the password. Authenticated REST calls should only be used via HTTPs.

However, I used `HTTPDigestAuth` to make sure and the answer from api.bintray.com is:

> {"message":"This resource requires authentication"}

So it's not supported.

SSE4: okay, this is definitely not good, but as we can't do anything with it anyway, leave it as is

hooks/conan-center.py

hooks/bintray_update.py
uilianries: Done

hooks/bintray_update.py
uilianries: Done

hooks/bintray_update.py
uilianries: Done

tox.ini
uilianries: Removed

uilianries: Ops! Seems like webtest is required because we dont install requirements_dev from Conan: https://travis-ci.org/conan-io/hooks/jobs/479512799#L256

hooks/bintray_update.py
uilianries: No, but we could check if is HTTPS and raise an error if not, WDYT?

SSE4: agree

uilianries: Done

uilianries:
Up! Should we wait for https://github.com/conan-io/hooks/pull/16 before to merge this PR?

jgsogo:
yes, #16 is the top priority now in this repo. I don't if #16 will get merged, but that discussion should be resolved before merging anything else (tests could be completely different)

jgsogo:
Hi @uilianries, we really want to start promoting hooks and this one can be very interesting. Although I think that testing will follow the approach in #17 (more or less like it is now) maybe it is better to split this PR into two: the hook itself (to be merged as soon as possible), and the testing part (so we can comment about it).

Also, please add a little documentation in the `README.md`, we are not sure if that will be the final place for documentation, but at least it will be the first place.

Thanks

uilianries:
Understood, I gonna do it today :smile_cat:

.travis.yml
```diff
@@ -98,7 +98,8 @@ install:
       pyenv activate conan;
     fi
   - pip install --upgrade pip
-  - pip install tox tox-venv tox-travis
+  - pip install -r tests/requirements_test.txt
+  - pip install -r .travis/requirements_travis.txt
```
jgsogo: Do you know that you can call one `requirements.txt` file from inside another? (https://pip.readthedocs.io/en/1.1/requirements.html#recursive-requirements)

So, you can have here just:

```bash
pip install -r .travis/requirements_travis.txt
```

and then add `-r tests/requirements_test.txt` (I don't know if the path should be relative to the `pip install` working dir or to the file).

README.md
```diff
@@ -34,11 +44,16 @@ If you handle multiple dependencies in your project is better to add a *conan.co
...
 
-### Attribute checker
+### [Attribute checker](hooks/attribute_checker.py)
 
 This hook checks that some important attributes are present in the ``ConanFile``: url, 
 license and description, and will output a warning for the missing ones. 
 
+### [Bintray Update](hooks/bintray_update.py)
+
+This hooks updates the Bintray package info by the attributes present in the ``ConanFile``.
```
jgsogo: Add something about the environment variables that this hook uses.

tests/utils/test_cases/conan_client.py
```diff
@@ -34,4 +34,4 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         os.chdir(cls._old_cwd)
-        shutil.rmtree(cls._working_dir)
\ No newline at end of file
+        shutil.rmtree(cls._working_dir, ignore_errors=True)
```
jgsogo: This modification shouldn't be needed.

README.md
uilianries: Done!

tests/utils/test_cases/conan_client.py
uilianries: Done!

.travis.yml
uilianries: Done!

tox.ini
jgsogo: Now that we are not importing anything from the `conans.tests` the `webtest` is no longer needed

uilianries: that's true! I forgot to remove it, thanks!

uilianries:
Thanks Javi!

uilianries:
@jgsogo  All green!

PR #4, opened 2018-10-19
WIP: binary linter

SSE4:
closes https://github.com/conan-io/conan/issues/1115
TODO: https://github.com/conan-io/conan/issues/1156 https://github.com/conan-io/conan/issues/1202

uses `LIEF` python module: https://github.com/lief-project/LIEF

- [x] verify executable type (PE on Windows, ELF on Linux, Mach-O on OSX)
- [x] verify executable architecture
- [x] verify compiler runtime for Visual Studio
- [x] verify no shared libraries when `shared=False`
- [ ] verify shared libraries actually export something
- [ ] verify `libstdc++` and `libc++` dependencies
- [ ] verify fPIC
- [ ] verify ELF endianess
- [ ] verify ELF ABI
- [x] verify dependencies on `cygwin.dll` or `msys.dll` for subsystems
- [ ] verify debug information for `Debug` and `RelWithDebInfo` builds
- [ ] verify Android API level
- [ ] verify OSX/iOS minimum version

limitations:
- [ ] LIEF doesn't currently support parsing of archives (static libraries), so no validation is done on them

questions:
- [ ] how to handle python module dependencies such as `LIEF` from plug-ins?

/cc @solvingj @Croydon @uilianries @lasote @memsharded @danimtb @jgsogo

### DIFF
--- a/hooks/binary-linter.py
+++ b/hooks/binary-linter.py
@@ -48,17 +48,12 @@ class BinaryLinter(object):
                 filename = os.path.join(root, filename)
                 self._verify_file(filename)
 
-    @property
-    def verbose(self):
-        return "VERBOSE" in os.environ
-
     def _verify_file(self, filename):
         self._filename = filename
         self._binary = lief.parse(filename)
 
         if not self._binary:
-            if self.verbose:
-                self._output.info('file "%s" is not a executable, skipping...' % filename)
+            self._output.info('file "%s" is not a executable, skipping...' % filename)
             return
 
         if self._binary.format != self._expected_format:
@@ -142,12 +137,12 @@ class BinaryLinter(object):
         if self._has_import(library):
             if not expected:
                 self._output.warn('"%s" imports library "%s"' % (self._filename, library))
-            elif self.verbose:
+            else:
                 self._output.info('"%s" imports library "%s"' % (self._filename, library))
         elif not self._has_import(library):
             if expected:
                 self._output.error('"%s" doesn\'t import library "%s"' % (self._filename, library))
-            elif self.verbose:
+            else:
                 self._output.info('"%s" doesn\'t import library "%s"' % (self._filename, library))
 
     @property


solvingj:
Really cool!  Great use case .

jgsogo:
Please, rebase this PR on the master branch to get CI support. Thanks!

SSE4:
I am going to update this one, add some additional checks and test, then I'll ping you
P.S. there is ongoing work in LIEF to support static libraries / archives / object files:
https://github.com/lief-project/LIEF/issues/233
until that, only shared libraries and executable will be checked

SSE4:
@jgsogo updated

hooks/binary-linter.py
```diff
@@ -48,12 +48,17 @@ def verify(self):
                 filename = os.path.join(root, filename)
                 self._verify_file(filename)
 
+    @property
+    def verbose(self):
+        return "VERBOSE" in os.environ
```
uilianries: Why not reuse `CONAN_LOGGING_LEVEL` to set the output level?

SSE4: like `CONAN_LOGGING_LEVEL <= 10`?

SSE4: I am not sure I get it. `ConanOutput` doesn't have `debug` logging method/level. how can I distinguish regular `info` messages and `debug` messages?

uilianries: nm, you are using `output`, not logger. I didn't read correct.

hooks/binary-linter.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
+
+import os
+import lief
```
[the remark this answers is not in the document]
SSE4: `pip install --upgrade lief`?

hooks/binary-linter.py
jgsogo: I think that you should use directly the `output.info`, `output.warning`, `output.error`,...  and it is Conan itself the one who has to control which level/messages show to the end user (although It is not implemented right now). I think it is better this approach than another env-var to document, and the logic being handled here.

SSE4: may we add `output.debug` then? because, outputting these messages always will be pretty noisy, I believe

hooks/binary-linter.py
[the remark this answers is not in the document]
uilianries: I think the point is: https://github.com/conan-io/hooks/issues/7#issuecomment-458505345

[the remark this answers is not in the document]
jgsogo: Maybe the `hook_manager` should handle and `ImportError` and print a proper message (wrapping every import in a try/except is overkill, although it is needed right now), what do you think, @danimtb?

hooks/binary-linter.py
jgsogo: Let people start using it, then we can introduce some conditionals in the hook (it should be really fast to release a new version), or introduce the output level handling in the Conan client (it would require a scheduled release).

hooks/binary-linter.py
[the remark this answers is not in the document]
danimtb: Could be something nice to have but I would say that we would need to use the hooks first to support the addition of the feature. Would go for the point here https://github.com/conan-io/hooks/issues/7#issuecomment-458505345 meanwhile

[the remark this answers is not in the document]
SSE4: so, leaving it as is for now?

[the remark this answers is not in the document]
jgsogo: It is OK 👍  https://github.com/conan-io/conan/issues/4433

SSE4:
yes, merge please, I'll add new checks one-by-one eventually

uilianries:
> I think it is worth to open an issue with the list of pending "tasks"

the eternal list of technical debt :sweat_smile:

jgsogo:
hahahaa, yes, it will be an issue opened forever, sure, but we can make a ✔️ from time to time.

PR #6, opened 2018-10-20
Update GitHub Info

SSE4:
similar to the Bintray Update plug-in

uses GitHub API https://developer.github.com/v3/

- [x] update description
- [x] update homepage (requires conan 1.9.0)
- [x] update topics/tags (requires conan 1.9.0)

depends on:
- https://github.com/conan-io/conan/issues/3671
- is there an issue for `homepage` attribute?


Croydon:
homepage was added in the same pull request and discussed in the same issue as the topics attribute

jgsogo:
We have added Travis and AppVeyor CI to this repo and an example test for the `attribute_checker` hook. We need to merge master into this PR (to get CI) and then implement some tests to check that the hook is working (it should be easy to mock `requests.*` function, but ping me if you need help).

Thanks!

SSE4:
I am going to update and test this one and then I will ping you, thanks

jgsogo:
Have a look at the discussion and the kind of testing we are talking about in https://github.com/conan-io/hooks/pull/2, feedback is very welcome.

jgsogo:
@SSE4, please move this _hook_ to the `hooks` folder, and document it in the `README.md`. We are willing to merge all the hooks and start promoting this feature.

Thanks

SSE4:
@jgsogo updated!

hooks/github-updater.py
```diff
@@ -0,0 +1,82 @@
...
+def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
+    del conanfile, reference, kwargs
+
+    github_token = os.getenv('GITHUB_TOKEN')
+    if not github_token:
+        output.info('no GITHUB_TOKEN environment variable is set, skipping GitHub updater')
+        return
+
+    conan_instance, _, _ = conan_api.Conan.factory()
+    recipe_info = conan_instance.inspect(path=conanfile_path, attributes=None)
```
danimtb: No need to do this as you already have the conanfile object available in this method

hooks/github-updater.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
```
danimtb: remove this

jgsogo:
Although I'm not a great fan of env-variables I think that another one is needed to allow the user to restrict the version/branch that will modify the Github properties, as they are unique for all the repository.

This env variable should be optional, and I'm not sure if it should be targetting the branch name or the recipe reference,... what do you think?

SSE4:
unfortunately these properties aren't per branch, but per repository, so how to do in this case?

jgsogo:
If I have a `CONAN_GITHUB_HOOK_MAIN_BRANCH=master` I can get the branch name from the repo, compare it to the value of that variable and make the request or not.

Or I can have a `CONAN_GITHUB_HOOK_MAIN_REF_PATTERN=*/*@conan/stable`, get the reference and act according to it.

Or maybe I should check if current branch is the default one in the remote repo (no need for an env variable).  

----

Maybe the option I like most is the last one...

hooks/github-updater.py
```diff
@@ -17,8 +16,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
         output.info('no GITHUB_TOKEN environment variable is set, skipping GitHub updater')
         return
 
-    conan_instance, _, _ = conan_api.Conan.factory()
-    recipe_info = conan_instance.inspect(path=conanfile_path, attributes=None)
+    recipe_info = conanfile.inspect(path=conanfile_path, attributes=None)
```
jgsogo: This won't work, I hope 😆 

You can access the data of the recipe using the `ConanFile` object you already have: `conanfile.url`, `conanfile.description`, `conanfile.homepage`,...

SSE4: okay, I'll revert it back to use `conan_api` then?

jgsogo: No, you already have the `ConanFile` object, no need to inspect, you have access to all the relevant information using the variable `conanfile`.

SSE4: okay, I'll try it. it turns out `ConanFile` object returned by `inspect` and `ConanFile` passed to the `pre_export` are different, right?

SSE4:
@jgsogo this might be useful, but I have some doubts:

- it adds dependency on Git executable availability
- it adds an additional complexity to the hook
- it requires package folder to be a git repository, which is not necessary a case (e.g. I may have downloaded an archive from GitHub)

I suppose better and cleaner approach is to just check conan channel if it's `stable` or `release`, WDYT?

jgsogo:
@SSE4, is this comment related to this PR?

SSE4:
@jgsogo yes, its about your suggestion for `CONAN_GITHUB_HOOK_MAIN_BRANCH`

jgsogo:
Ok, I forgot my previous comment, I was just checking the sources 😓 

Yes, the suggestion about checking against a local repo is stupid, forget it... nevertheless, I would like to know more opinions about the second one (or something similar), two ways to implement it:

 * CONAN_GITHUB_HOOK_MAIN_REF_PATTERN=*/*@conan/stable
 * Query Github to know which one is the _default_ branch (I think it should always be the last stable version, isn't it?) and update that information only if it matches the version/channel of the Conan reference.

SSE4:
the last one is the most logical and simple for me (query GitHub default branch)

jgsogo:
Also, it doesn't require another env variable... and I really think it is useful to keep the information updated between the repo and the last version of the recipe.

SSE4:
@jgsogo I've added required check for the default branch

PR #16, opened 2019-01-10
New module structure, testing approach and example in conan-center

danimtb:
Here I'd like to introduce a new approach to developing functionalities for hooks.

I have created a new module "checks" (not definitive name at all) that gathers the functions that performs the checks or any other action that could be done in a hook. this way I think we can split the functionalities to reused in different hooks and DRY.

I have also created a new structure similar to the conan codebase with the same organization for tests. I think having a similar way of working and organizing things is better for contributors.

Finally, I included some mock classes needed for the tests. They were extracted from the Conan test suite and although doing the job right now, they are still in a preliminar phase.

Would appreciate some feedback following the discussion in #2 @jgsogo @uilianries


uilianries:
I really liked the idea about DRY! :clap:

jgsogo:
My opinion is against the layout changes, although it shares some parallelism with Conan main repo:

 - `tests` should be a different folder with all the code related to tests (also _utils_ that are built just for the tests themselves). This has two advantages:
   - when creating a package for pypi, that folder will be omitted. Otherwise the package will contain also the tests, it results in a bigger package to be uploaded/downloaded (it is happening now with Conan)
   -  you avoid people importing things from your tests classes/helpers and so on. Here we were tempted to do a `from conans.tests.... import ...`, if the `tests` folder is outside the repo one, this won't be possible.
 - No `__init__.py` in the root directory. Python code starts inside `hooks` and `tests`.
 - I will go only for the `unittests` approach, as hooks should be small pices of functionality, not related one with another, I think there is no motivation for a `tests/functional` module. We can have:
   - `tests/<common-functionality>/test_....`
   - `tests/<hook-name>/test_....`
   - `tests/¿consistency?/...`: these will be related to the helper classes (some of them introduced in this PR) and this test suite should be the only one that runs against all the Conan versions.

We should agree on this before moving forwards

danimtb:
Agree with the comments. 

The only point I don't understand is:

> tests/¿consistency?/...: these will be related to the helper classes (some of them introduced in this PR) and this test suite should be the only one that runs against all the Conan versions.

Could you please elaborate? Only unit tests or run against different Conan versions?

jgsogo:
Elaborating 👍 : 

In the future, the number of hooks will grow up and unittests will take longer to run, we will like them to be as fast as possible and a handicap will be to run those unittests against every Conan version.

If we want the unittest to run using only one Conan version we should check that the inputs and outputs for the hooks are supported/consistent for every Conan version. It would be enough to run some _consistency_ checks:
 * check input argument for every _hook_ function: number, naming and duck-type (I'm not sure if type will be needed too).
 * check that our mocked arguments in the test suite behave like the real classes that will be feeding the hooks

If we go for the unittest approach, this is the only way I see it can work; otherwise we should keep the previous approach where the hooks where installed, Conan was run and mocking should use other strategies to intercept the workflow.... or take the _Conan API_ path.

jgsogo:
Maybe we have to rewrite this PR from scratch...

PR #17, opened 2019-01-23
Run tests within the same python process

jgsogo:
Hi! I want to share with you this _trick_ to run the tests, I'm calling the Conan `main` function (not actually because it contains some `sys.exit` calls in case of failure) to run the commands.

We think this is almost the best approach to run the tests:
 * hooks are actually tested using actual Conan flow
 * we run the code for several Conan versions, we are checking not only the compatibility regarding the input parameters, but also the tools that could be used inside the hooks.
 * we can mock whatever we want, we are inside the same python process 💯 

Let me know your thoughts about it.

### DIFF
--- a/tests/utils/test_cases/conan_client.py
+++ b/tests/utils/test_cases/conan_client.py
@@ -2,11 +2,26 @@
 
 import os
 import shutil
-import subprocess
 import tempfile
 import unittest
+from io import StringIO
 
+from conans.client.command import Conan, CommandOutputer, Command, SUCCESS
 from tests.utils.environ_vars import context_env
+
+
+class OutputStream(object):
+    def __init__(self):
+        self._data = StringIO()
+
+    def write(self, data):
+        self._data.write(data)
+
+    def flush(self):
+        pass
+
+    def as_str(self):
+        return self._data.getvalue()
 
 
 class ConanClientTestCase(unittest.TestCase):
@@ -19,11 +34,18 @@ class ConanClientTestCase(unittest.TestCase):
         kwargs.update({'CONAN_USER_HOME': home, 'CONAN_USER_HOME_SHORT': home_short})
         return kwargs
 
-    def conan(self, *command, **kwargs):
+    def conan(self, *command, expected_return_code=SUCCESS):
         with context_env(**self._get_environ()):
-            p = subprocess.Popen(*command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            out, err = p.communicate()
-            return out.decode()
+            # This snippet reproduces code from conans.client.command.main, we cannot directly
+            # use it because in case of error it is exiting the python interpreter :/
+            conan_api, cache, user_io = Conan.factory()
+            output_stream = OutputStream()
+            user_io.out._stream = output_stream
+            outputer = CommandOutputer(user_io, cache)
+            cmd = Command(conan_api, cache, user_io, outputer)
+            return_code = cmd.run(*command)
+            self.assertEqual(return_code, expected_return_code)
+            return output_stream.as_str()
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


tests/utils/test_cases/conan_client.py
```diff
@@ -19,11 +21,27 @@ def _get_environ(self, **kwargs):
...
+                return_code = cmd.run(command)
+            except Exception as e:
+                # Conan execution failed
+                self.fail("Conan execution for this test failed with {}: {}".format(type(e), e))
+            else:
+                # Check return code
+                self.assertEqual(return_code, expected_return_code,
+                                 msg="Unexpected return code\n\n{}".format(output_stream.getvalue()))
+            finally:
+                conan_api._remote_manager._auth_manager._localdb.connection.close() # Close sqlite3
```
jgsogo: Conan is not closing the connection to the database 😮 , we need to close it in order to remove the working directory.

uilianries:
Sounds like integration tests.

tests/utils/test_cases/conan_client.py
```diff
@@ -34,4 +52,4 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         os.chdir(cls._old_cwd)
-        shutil.rmtree(cls._working_dir)
\ No newline at end of file
+        shutil.rmtree(cls._working_dir)
```
uilianries: I saw random errors on Windows. Should we ignore the errors when removing the working dir?

tests/utils/test_cases/conan_client.py
```diff
@@ -2,10 +2,12 @@
 
 import os
 import shutil
-import subprocess
+import stat
```
uilianries: I didn't find any usage

uilianries: fixed

tests/utils/test_cases/conan_client.py
jgsogo: What kind of random errors? Can you link to any?

uilianries: most related to busy working dir on Windows, same from Conan tests. I saw those errors on Conan's Jenkins jobs.

but as you are using this statement for tests only, I see no problem here.

jgsogo: I thought you mean errors/warnings in this repo... for Conan itself this will always raise because the connection to the database is never closed (here I'm closing it), and maybe some tools may break in the future if they leave some file open or so, but I think we should keep this here and we will see what to do if something fails

uilianries: I agree.

tests/utils/test_cases/conan_client.py
```diff
@@ -19,11 +22,27 @@ def _get_environ(self, **kwargs):
...
+                return_code = cmd.run(command)
+            except Exception as e:
+                # Conan execution failed
+                self.fail("Conan execution for this test failed with {}: {}".format(type(e), e))
+            else:
+                # Check return code
+                self.assertEqual(return_code, expected_return_code,
+                                 msg="Unexpected return code\n\n{}".format(output_stream.getvalue()))
+            finally:
+                conan_api._remote_manager._auth_manager._localdb.connection.close()  # Close sqlite3
```
danimtb: Deserves an issue in the Conan repo. Don't you think?

danimtb:
Quite smooth and seems the way to go! 👍

tests/utils/test_cases/conan_client.py
jgsogo: Yes, open it, although this line must be here (maybe conditional) to be able to test previous versions.

danimtb: Done: https://github.com/conan-io/conan/issues/4400

PR #20, opened 2019-01-31
Enhancements to test suite

jgsogo:
Tox is running the test-suite, so it should be that utility the one that installs the requirements. Also, I'm adding a couple of utilities that I will be using in my next PR (please, be gentle with those additions)

### DIFF
--- a/tests/utils/test_cases/conan_client.py
+++ b/tests/utils/test_cases/conan_client.py
@@ -15,7 +15,6 @@ class ConanClientTestCase(unittest.TestCase):
     """ Helper class to run isolated conan commands """
     _old_cwd = None
     _working_dir = None
-    _user_subdir = 'dev'
 
     def _get_environ(self, **kwargs):
         # kwargs = super(ConanClientTestCase, **kwargs)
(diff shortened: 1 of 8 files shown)


tests/utils/test_cases/conan_client.py
```diff
@@ -47,13 +46,9 @@ def conan(self, command, expected_return_code=SUCCESS):
...
 
     def setUp(self):
-        user_working_dir = os.path.join(self._working_dir, self._user_subdir)
-        os.makedirs(user_working_dir)
-        os.chdir(user_working_dir)
-
-    def tearDown(self):
-        # Clean working directory
-        shutil.rmtree(os.path.join(self._working_dir, self._user_subdir))
+        testcase_dir = os.path.join(self._working_dir, str(uuid.uuid4()))
```
uilianries: why not `tempfile.mkdtemp` ?

jgsogo: I'm creating all the temp folders inside the `self._working_dir` and collecting the garbage after each test case in the `tearDownClass` function.

PR #39, opened 2019-02-21
 Review conan-center hook

### LINKED ISSUE #37: [conan-center] settings.os should not be mandatory
Hi!

When building an installer package (bincrafters/pkg-config_installer), only the settings `os_build` and `arch_build` are available. However, the hook `conan-center` requires `settings.os` as well:

```
ERROR: [HOOK - conan-center.py] post_build(): 'settings.os' doesn't exist
'settings' possible configurations are ['arch_build', 'compiler', 'os_build']
Traceback (most recent call last):
  File "/opt/python/pyenv/versions/3.6.0/lib/python3.6/site-packages/conans/client/hook_manager.py", line 56, in execute
    method(output, **kwargs)
  File "/home/uilian/.conan/hooks/conan-center.py", line 121, in post_build
    if not _files_match_settings(conanfile, conanfile.build_folder):
  File "/home/uilian/.conan/hooks/conan-center.py", line 209, in _files_match_settings
    os = _get_os(conanfile)
  File "/home/uilian/.conan/hooks/conan-center.py", line 243, in _get_os
    if hasattr(conanfile.settings, attrib):
  File "/opt/python/pyenv/versions/3.6.0/lib/python3.6/site-packages/conans/model/settings.py", line 257, in __getattr__
    self._check_field(field)
  File "/opt/python/pyenv/versions/3.6.0/lib/python3.6/site-packages/conans/model/settings.py", line 253, in _check_field
    raise undefined_field(self._name, field, self.fields, self._parent_value)
conans.errors.ConanException: 'settings.os' doesn't exist
'settings' possible configurations are ['arch_build', 'compiler', 'os_build']
```

We should check if `os` or `os_build` are available.

Regards!

danimtb:
fixes #37 

- Review issues of the hook
- Added basic tests
- Included documentation

### DIFF
--- a/tests/utils/test_cases/conan_client.py
+++ b/tests/utils/test_cases/conan_client.py
@@ -7,7 +7,9 @@ import unittest
 import uuid
 from io import StringIO
 
+from conans.__init__ import __version__ as conan_version
 from conans.client.command import Conan, CommandOutputer, Command, SUCCESS
+from conans.model.version import Version
 from tests.utils.environ_vars import context_env
 
 
@@ -42,7 +44,8 @@ class ConanClientTestCase(unittest.TestCase):
                 self.assertEqual(return_code, expected_return_code,
                                  msg="Unexpected return code\n\n{}".format(output_stream.getvalue()))
             finally:
-                conan_api._remote_manager._auth_manager._localdb.connection.close()  # Close sqlite3
+                if Version(conan_version) < "1.13":
+                    conan_api._remote_manager._auth_manager._localdb.connection.close()  # Close sqlite3
             return output_stream.getvalue()
 
     def setUp(self):
--- a/README.md
+++ b/README.md
@@ -44,6 +44,31 @@ If you handle multiple dependencies in your project is better to add a *conan.co
 
 These are the hooks currently available in this repository
 
+### [Conan Center](hooks/conan-center.py)
+
+This hook does checks for the [inclusion guidelines of third-party libraries](https://docs.conan.io/en/latest/uploading_packages/bintray/conan_center_guide.html#inclusion-guidelines-for-third-party-libraries)
+in [Conan Center](https://bintray.com/conan/conan-center).
+
+It is mostly intended for users who want to contribute packages to Conan Center. With this hook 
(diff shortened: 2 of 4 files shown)


danimtb:
I have included a version condition in the close of the Conan database that was fixed in the Conan 1.13 develop branch

tests/utils/test_cases/conan_client.py
```diff
@@ -42,7 +44,8 @@ def conan(self, command, expected_return_code=SUCCESS):
                 self.assertEqual(return_code, expected_return_code,
                                  msg="Unexpected return code\n\n{}".format(output_stream.getvalue()))
             finally:
-                conan_api._remote_manager._auth_manager._localdb.connection.close()  # Close sqlite3
+                if Version(conan_version) < "1.13":
+                    conan_api._remote_manager._auth_manager._localdb.connection.close()  # Close sqlite3
```
jgsogo: Just curious: was it failing in the test suite against the develop branch?

danimtb: Yes, you can see it here: https://travis-ci.org/conan-io/hooks/jobs/496412258#L292

danimtb:
Ok, will try to check always the ConanFile class. I will take a look before merging.

I know that about ``no_copy_source``, but the check means that it is kind of mandatory for header only files. For other packages it is something optional and therefor you cannot check if.

Regarding the last comment, I don't know where that link is pointing to :neutral_face:

jgsogo:
The link... it appears with the yellow line, but it doesn't take you to that line :/ It is in the function `def _has_files_with_extensions(folder, extensions):`, there you keep iterating files although the returning value is already `True`.

tests/utils/test_cases/conan_client.py
jgsogo: I was asking if it was failing because of the double close, not because of that attribute being no longer there. 

👍

danimtb:
Please check when possible. Travis CI passed although the icon is red (it points to a different build 😕)

jgsogo:
There are still several uses of the parsed conanfile text... I'm sure we will have to work in this hook in the future, but it can be enough for a first ride.

PR #47, opened 2019-02-25
#36 Bintray hooks should use Bintray variables

### LINKED ISSUE #36: Credentials of bintray plugin env vars
About these variables:
```
CONAN_LOGIN_USERNAME: Bintray login username
CONAN_PASSWORD: Bintray API KEY
```
I know that these variables are used in CPT, but even these I would like to rename them.
I think would make more sense to name them as "BINTRAY_XXX" even check if there are some "standard" variables maybe in the jfrog cli?

uilianries:
Hi!

This PR renames the Bintray account variables, like username and password.

fixes #36 
/cc @lasote

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ license and description, and will output a warning for the missing ones.
 This Conan hook reads your recipe and updates its Bintray package info using the attributes.
 
 It's necessary pass Bintray login by environment variables:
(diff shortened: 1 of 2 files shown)


danimtb:
CI looking good (nothing broken) - In fact there are no tests for this hook yet, so... -
Merging this!

PR #54, opened 2019-03-01
Verbose Github Updater (#43)

### LINKED ISSUE #43: [github-updater] Increase output information
The github-updater hook works silently and in my case it only outputs:

```
[HOOK - conanio/hooks/github-updater.py] pre_export(): no topics to update
```

I don't know if it is done on purpose, but it should say what this it is updating in github.

uilianries:
Hi!

This PR is not just about #43, but also it includes all tests for github-updater. The coverage is 100% for that hook.

fixes #43


hooks/github_updater.py
```diff
@@ -1,87 +1,165 @@
...
-    if not github_token:
-        output.info('no GITHUB_TOKEN environment variable is set, skipping GitHub updater')
-        return
+The hook is automatically called when export command is executed:
 
+    $ conan export . package/0.1.0@user/channel
+    [HOOK - github_updater.py] pre_export(): The attributes are up-to-date.
+    [HOOK - github_updater.py] pre_export(): The topics are up-to-date.
+    Exporting package recipe
+"""
```
danimtb: Is this PEP8 compliant? No blank line needed?

tests/test_hooks/test_github_updater.py
```diff
@@ -0,0 +1,296 @@
...
+import os
+import textwrap
+import responses
+
+from conans import tools
+
+from tests.utils.test_cases.conan_client import ConanClientTestCase
+
+
+_GITHUB_REPO_DATA = {
```
danimtb: Can we use here only the relevant information for this hook?

uilianries: the hook uses only 3 attributes from this response. This response follow the Github v3 format. I gonna reduce it, because I think there is problem extra information, since we don't treat any other field.

hooks/github_updater.py
uilianries: good question. I used yapf to format my file, but blank line would be nice. I'll add it.

uilianries: done!

tests/test_hooks/test_github_updater.py
uilianries: done!

danimtb: Much better! Thank you

PR #55, opened 2019-03-01
Improve error message (#41)

### LINKED ISSUE #41: [bintray-update] Wont work with conan_server
Hi!

When running conan_server, the hooks bintray-update shown:

```
[HOOK - bintray_update.py] post_upload_recipe(): ERROR: Could not extract subject and repo from http://0.0.0.0:9300: Invalid pattern
```

We need to check before to extract the information if the endpoint is Bintray.

uilianries:
Hi!

The message reported on #41 is not clear enough, because only a pattern is used to compare the remote address, but there is no detail about. I think the user only wants to know if the remote is valid to be updated.

fixes #41

### DIFF
--- a/hooks/bintray_updater.py
+++ b/hooks/bintray_updater.py
@@ -65,7 +65,7 @@ def _extract_user_repo(remote):
     pattern = r'https?:\/\/api.bintray.com\/conan\/(.*)\/(.*)'
     match = re.match(pattern=pattern, string=remote.url)
     if not match:
-        raise ValueError("Could not extract subject and repo from %s: Invalid pattern" % remote.url)
+        raise ValueError("The remote {} is not a Bintray repository." % remote.name)
     return match.group(1), match.group(2)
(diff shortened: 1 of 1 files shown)


danimtb:
Does that exception makes the conan client finish with error? (sorry, I can't remember)
I think it shouldn't

uilianries:
> Does that exception makes the conan client finish with error? (sorry, I can't remember)
I think it shouldn't

Yes, but I have used try..except to print an error message only (see line [55](https://github.com/conan-io/hooks/blob/f4c54867ec0c616e026ab9831cf39618bb9a7955/hooks/bintray_updater.py#L55), thus there is no exception when running this hook.

IMO if this hook is active it's because someone wants to update the Bintray page, so I prefer consider any warning as error, Otherwise, the user could deactivate the hook.

danimtb:
Great, thanks for the clarification!

PR #56, opened 2019-03-01
Upload only supported licenses (#38)

### LINKED ISSUE #19: Create tests for Bintray plugin
Related issue: https://github.com/conan-io/hooks/pull/2

Use same approach from https://github.com/conan-io/hooks/pull/17 to design the tests.

### LINKED ISSUE #38: [bintray-update] Map SPDX licenses
Hi!

There is a lack between Bintray and SPDX licenses, where Bintray only supports a limited number of licenses, but any Conan recipe could have one or multiple licenses, following the SPDX format.

This feature intents to map SPDX licenses to Bintray format, fixing possible lacks.
For example:

SPDX GPL-2.0-or-later -> Bintray GPL-2.0

Regards!

uilianries:
Hi!

- Only upload license supported by Bintray
- Fix BSD license names
- Add Hook Tests

fixes #38
fixes #19


uilianries:
Ready to review!

/cc @jgsogo @danimtb @SSE4

hooks/bintray_updater.py
```diff
@@ -196,12 +213,13 @@ def _get_branch():
...
-    for ci, branch in ci_manager.items():
-        if os.getenv(ci) and os.getenv(branch):
+    for manager, branch in ci_manager.items():
+        if os.getenv(manager) and os.getenv(branch):
             return os.getenv(branch)
 
     try:
-        for line in subprocess.check_output("git branch --no-color", shell=True).decode().splitlines():
+        for line in subprocess.check_output(
+                "git branch --no-color", shell=True).decode().splitlines():
```
danimtb: use ``tools.Git()`` ?

hooks/bintray_updater.py
```diff
@@ -25,13 +25,12 @@
...
 
-
 __version__ = '0.1.0'
 __license__ = 'MIT'
-__author__  = 'Conan.io <https://github.com/conan-io>'
+__author__ = 'Conan.io <https://github.com/conan-io>'
 
 
-def post_upload_recipe(output, conanfile_path, reference, remote, **kwargs):
+def pre_upload_recipe(output, conanfile_path, reference, remote, **kwargs):
```
danimtb: What happens if the package entry is not created in BIntray yet?

hooks/bintray_updater.py
uilianries: I forgot again about this tool! thanks!

hooks/bintray_updater.py
uilianries: You are right! That's why I used `post_upload_recipe`. I gonna revert it  because it will fail if there is no package there. Thanks!

uilianries: Done!

hooks/bintray_updater.py
uilianries: Done!

uilianries:
Seems like Bintray changed something on API: https://stackoverflow.com/questions/55365144/bintray-api-can-not-obtain-oss-licenses-as-anonymous

That error will affect this hook. Also, I sent an email to JFrog support. I'll post the answer here, when available.

uilianries:
I think this should be enough. Since the API is authenticated, we can retrieve the OSS licenses.

@danimtb @SSE4 @jgsogo please review when possible.

tests/utils/licenses.py
```diff
@@ -0,0 +1,486 @@
+# -*- coding: utf-8 -*-
+
+# Bintray OSS Licenses
+# https://api.bintray.com/licenses/oss_licenses
```
jgsogo: If it is just for testing, and this URL is not public, then it would be better to add just a couple of examples in the list, no need to expose/maintain a full list, don't you think so?

uilianries: @jgsogo Done!

grv87:
Guys, there are much more differences between Bintray and SPDX beyond just `BSD`. Consider e.g. `AGPL-V3` vs `AGPL-3.0-only`/`AGPL-3.0-or-later` or `Mozilla-1.1` vs `MPL-1.1`.

uilianries:
@grv87 I totally forgot about AGPL, thanks!

danimtb:
Now with conan-io/conan-center-index, the usage of this hook would be reduced. However, as this PR has some work on it, we could complete it with the latest comments.

uilianries:
@danimtb what about consuming the list from https://spdx.org/licenses/ ?

We can print a warning when a license is not available on Bintray list.

markgalpin:
You may find: https://github.com/jfrog/gocenter/blob/master/acceptedLicenses.json a useful reference...

uilianries:
> You may find: [jfrog/gocenter:acceptedLicenses.json@`master`](https://github.com/jfrog/gocenter/blob/master/acceptedLicenses.json) [](/jfrog/gocenter/blob/HEAD@{2019-10-03T21:43:52Z}/acceptedLicenses.json) a useful reference...

That's help a lot! Thanks @markgalpin !

jgsogo:
Also, it looks like that the endpoint https://api.bintray.com/licenses/oss_licenses doesn't require authentication right now. There we can find the actual list of licenses supported by Bintray.

I think we can rebase this PR on top of `master` and reduce changes to a minimum, and we would be able to review and add it to the set of hooks.

tests/test_hooks/test_bintray_updater.py
```diff
@@ -0,0 +1,323 @@
...
+        """
+        responses.replace(responses.GET, "https://api.bintray.com/licenses/oss_licenses", status=500, json={"message": "You have reached a dark spot"})
+        tools.save('conanfile.py', content=self.conanfile_complete)
+        self.conan(['export', '.', 'dummy/0.1.0@foobar/stable'])
+        output = self.conan(['upload', '--remote=fake', 'dummy/0.1.0@foobar/stable'])
+        self.assertIn("Uploading dummy/0.1.0@foobar/stable to remote 'fake'", output)
+        self.assertIn("Uploaded conan recipe 'dummy/0.1.0@foobar/stable' to 'fake'", output)
+        self.assertIn("post_upload_recipe(): Reading package info from Bintray.", output)
+        self.assertIn("post_upload_recipe(): Inspecting recipe info.", output)
+        self.assertIn('post_upload_recipe(): ERROR: Could not request OSS licenses (500): {"message": "You have reached a dark spot"}', output)
```
jgsogo: Shall we add more info to the user? Something like: "this hook requires access to OSS license list in Bintray (URL), you may deactivate the hook if the situation persists"... or we could provide a workaround: an environment variable to get the OSS licenses list from a local JSON file.

tests/test_hooks/test_bintray_updater.py
```diff
@@ -0,0 +1,323 @@
...
+        responses.add(responses.POST, 'https://api.fake.io/v1/conans/dummy/0.1.0/foobar/stable/upload_urls', json={})
+        self.conan(['remote', 'add', 'badremote', 'https://api.fake.io'])
+        tools.save('conanfile.py', content=self.conanfile_complete)
+        self.conan(['export', '.', 'dummy/0.1.0@foobar/stable'])
+        output = self.conan(['upload', '--remote=badremote', 'dummy/0.1.0@foobar/stable'])
+        self.assertIn("Uploading dummy/0.1.0@foobar/stable to remote 'badremote'", output)
+        self.assertIn("Uploaded conan recipe 'dummy/0.1.0@foobar/stable' to 'badremote'", output)
+        self.assertNotIn("post_upload_recipe(): Reading package info from Bintray.", output)
+        self.assertNotIn("post_upload_recipe(): Inspecting recipe info.", output)
+        self.assertIn("post_upload_recipe(): ERROR: The remote 'badremote' is not a valid Bintray URL.", output)
```
jgsogo: Again, if I work with several remotes, this hook cannot fail when I upload a package to a non-bintray remote. I think it is too strict. I don't know which one is the best solution, maybe: if it is not a bintray URL, warn and do nothing in this hook (wdyt, @danimtb?).

tests/test_hooks/test_bintray_updater.py
```diff
@@ -0,0 +1,323 @@
...
+            The hook must not continue when is not possible to retrieve the cached information.
+        """
+        responses.replace(responses.GET, 'https://api.bintray.com/packages/foobar/conan/dummy%3Afoobar', status=500, json={"message": "You have reached a dark spot"})
+        tools.save('conanfile.py', content=self.conanfile_complete)
+        self.conan(['export', '.', 'dummy/0.1.0@foobar/stable'])
+        output = self.conan(['upload', '--remote=fake', 'dummy/0.1.0@foobar/stable'])
+        self.assertIn("Uploading dummy/0.1.0@foobar/stable to remote 'fake'", output)
+        self.assertIn("Uploaded conan recipe 'dummy/0.1.0@foobar/stable' to 'fake': https://bintray.com/foobar/conan", output)
+        self.assertIn("post_upload_recipe(): Reading package info from Bintray.", output)
+        self.assertIn('ERROR: Could not request package info (500): {"message": "You have reached a dark spot"}', output)
```
jgsogo: I assume this doesn't fail if the package is not already in the remote... right?

hooks/bintray_updater.py
```diff
@@ -40,16 +44,23 @@ def post_upload_recipe(output, conanfile_path, reference, remote, **kwargs):
...
+        output.info("Inspecting recipe info.")
         recipe_info = _get_package_info_from_recipe(conanfile_path=conanfile_path)
-        updated_info = _update_package_info(recipe_info=recipe_info, remote_info=remote_info)
+        supported_licenses = _get_oss_licenses()
+        updated_info = _update_package_info(recipe_info=recipe_info, remote_info=remote_info,
+                                            supported_licenses=supported_licenses)
         if updated_info:
-            output.info("Bintray is outdated. Updating Bintray package info: {}".format(" ".join(updated_info.keys())))
-            _patch_bintray_package_info(package_url=package_url, package_info=updated_info, remote=remote)
+            output.info("Bintray is outdated. Updating Bintray package info: {}.".format(" ".join(
```
jgsogo:
```suggestion
            output.info("Bintray is outdated. Updating Bintray package info: '{}'.".format("', '".join(
```

hooks/bintray_updater.py
```diff
@@ -104,55 +114,74 @@ def _get_package_info_from_recipe(conanfile_path):
...
-        if isinstance(licenses, str):
-            licenses = [licenses]
-        if not bool(set(licenses).intersection(remote_info['licenses'])):
+        # INFO (uilianries): Bintray does not follow SDPX for BSD licenses
+        for version in [2, 3]:
+            licenses = [
+                "BSD %d-Clause" % version if it.lower() == ("bsd-%d-clause" % version) else it
+                for it in licenses
+            ]
+        licenses = [it for it in licenses if it in supported_licenses]
```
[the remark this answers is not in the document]
uilianries: That's it. Any license which is not supported by Bintray will generate an error when publishing, that's why I'm ignoring. 

But, yes, a warning message sounds good.

tests/test_hooks/test_bintray_updater.py
uilianries: A Json file sounds good, but as a hook file is only an independent python file, we would need create a dictionary actually. Anyway, I liked your message, I gonna update it.

tests/test_hooks/test_bintray_updater.py
uilianries: So do you prefer skipping this hook when pushing to Artifactory and conan_server?

tests/test_hooks/test_bintray_updater.py
uilianries: it can fail because we need to get the current information from remote to compare if we need to update.

hooks/bintray_updater.py
uilianries: done

tests/test_hooks/test_bintray_updater.py
uilianries: I've update the message error. please, review again.

hooks/bintray_updater.py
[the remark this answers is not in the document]
uilianries: I've updated the warning message. please, review again.

tests/test_hooks/test_bintray_updater.py
jgsogo: but it cannot fail if the package doesn't exist...

uilianries: I didn't get ... how can we know if the package exists when the server is down? I understand we shouldn't emit an error when a package does exist, but this test validates a server error.

uilianries:
Seems like Python 2.7 is not working with Conan development version. @jgsogo should we drop that test?

jgsogo:
Python 2.7 works, it is just a typical unicode problem. I think you can use `{!r}` to format the string or probably a `str(e)` should work.

tests/test_hooks/test_bintray_updater.py
jgsogo: Sorry, I didn't realize it was a 500 error

jgsogo:
@danimtb, please, I would want to know your opinion [this comment](https://github.com/conan-io/hooks/pull/56/files/7e26d17db9569e56fd0afd81d114eb039bddbddb#diff-4cda5e633c79e356fab680b73943008f). How can we deal with it?

tests/test_hooks/test_bintray_updater.py
danimtb: It would be nice if the hook could automatically exclude remotes that are not Bintray, but it is an enhancement we can postpone for a future update

PR #57, opened 2019-03-05
Add tests for hook loader: colliding files and modules

jgsogo:
This PR adds some tests trying to demonstrate what happens if several hooks try to import modules/files with the same name.

It is related to the PR proposed here: https://github.com/conan-io/conan/pull/4641 and https://github.com/conan-io/conan/pull/4664. Nevertheless, the loading function is not the same for the `HookManager` and the `conanfile.py`.

Fails here are not solvable in the hooks side.


jgsogo:
PR merged in conan repo. No longer needed.

PR #62, opened 2019-04-05
WIP: Conan center hook

lasote:
- Renamed to conan-center.
- Simplified code by using decorators managing the output
- Introduced a way to fail when an error, by an environment variable.
- To discuss if the decorators might have sense in other generators (provide them from conan?).
- Tested with c3i


README.md
```diff
@@ -24,6 +24,11 @@ To install all hooks from Conan repository in Github:
 
 ``$ conan config install https://github.com/conan-io/hooks``
 
+If you are using Conan >=1.14 you can specify the source and destination folder to avoid copying
```
uilianries: nice!

hooks/conan-center.py
```diff
@@ -0,0 +1,309 @@
+import fnmatch
```
uilianries: you should have used `git mv` :sweat:

hooks/conan-center.py
```diff
@@ -0,0 +1,309 @@
...
+    return tmp
+
+
+@raise_if_error_output
+def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
+    conanfile_content = tools.load(conanfile_path)
+    settings = getattr(conanfile, "settings", None)
+
+    @run_test("REFERENCE LOWERCASE", output)
+    def test(out):
```
uilianries: is it correct, all methods have the same name `test`

lasote: the name is irrelevant

hooks/conan-center.py
lasote: not sure if even using git mv it works if the diff is > X%

hooks/conan-center.py
uilianries: I see you are using a wrapper, but my eyes hurt when I see duplicate names :joy_cat:

hooks/conan-center.py
uilianries: Yes, it's a totally new hook :heart_eyes:

hooks/conan-center.py
jgsogo: I like the implementation, it's cleaner, but I would make each `test` a free function and then call them from hook's methods, I think it is better that implicitly calling the decorated function from the decorator itself (not tested, but I have some doubts about it):

```python
def reference_lowercase(output, reference, **kwargs)
    ...

@raise_if_error_output
def pre_export(*args, **kwargs):
    run_test("REFERENCE LOWERCASE", reference_lowercase, **kwargs)
```

danimtb:
A fix for the tests already in place would be needed

lasote:
I don't want to merge it yet actually. we have to validate if it covers every rule or restriction we want for the new ccci

danimtb:
Please update this PR with the changes in Master branch so both Travis and Appveyor CIs pass the tests. We fixed some issues last week.

danimtb:
Now there are some tests failing due to new checks introduced recently

PR #69, opened 2019-07-31
Up Conan versions

jgsogo:
From time to time we need to update the versions this repo is using to test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,39 +29,39 @@ jobs:
       dist: xenial
       sudo: true
 
-    - stage: Conan 1.15 - Linux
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -29,39 +29,39 @@ jobs:
       dist: xenial
       sudo: true
 
-    - stage: Conan 1.15 - Linux
+    - stage: Conan 1.18 - Linux
```
uilianries: It's duplicated with the current version, but it's good for me, since we have patch version frequently.

jgsogo: You are right, but you have to think about the future 😅

PR #70, opened 2019-08-06
feat: add support for uploading Conan Python package to a PyPI index

martinm82:


lasote:
Hi! I don't really understand the use case, but at first sight, a post upload hook that publishes to pypi sounds a bit weird. How does it know which packages have to be uploaded to pypi? How are you using this generator?

PR #72, opened 2019-09-04
fixup install command in README.md

ericLemanissier:
without the .git extension, conan expects an archive file


danimtb:
oops, didn0't notice that. Thanks for the fix!

PR #80, opened 2019-09-27
fPIC should not always be removed

uilianries:
The Conan Center hooks, validates when fPIC is declared in the recipe, but also requires that fPIC must be removed for Windows. However, some project are not supported on Linux (e.g. libmount), so there is no need for such action.

This PR runs `conan info` to validate the recipe support for Windows. If ConanInvalidConfiguration is raise, it's because the recipe can't be built on Windows, so we don't need to remove fPIC from the recipe. 

Related PR: https://github.com/conan-io/conan-center-index/pull/123

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -245,7 +245,7 @@ def pre_build(output, conanfile, **kwargs):
         has_fpic = conanfile.options.get_safe("fPIC")
         if conanfile.settings.get_safe("os") == "Windows" and has_fpic:
             out.error("'fPIC' option not managed correctly. Please remove it for Windows "
-                        "configurations: del self.options.fpic")
+                      "configurations: del self.options.fpic")
         elif has_fpic:
             out.success("OK. 'fPIC' option found and apparently well managed")
         else:
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -229,34 +229,20 @@ def post_source(output, conanfile, conanfile_path, **kwargs):
                 out.error(
                         "Can't detect C++ source files but recipe does not remove 'compiler.libcxx'")
 
+
+@raise_if_error_output
+def pre_build(output, conanfile, **kwargs):
+
     @run_test("KB-H007", output)
     def test(out):
-        conanfile_content = tools.load(conanfile_path)
-        low = conanfile_content.lower()
-        options = getattr(conanfile, "options")
-        options = options if options is not None else []  # options = None by default
-        if "fPIC" in options:
-            remove_fpic_option = ["self.options.remove(\"fpic\")",
-                                  "self.options.remove('fpic')",
-                                  "del self.options.fpic"]
-            os_compare = ['self.settings.os == "windows"',
-                          "self.settings.os == 'windows'",
-                          'self.settings.os != "linux"'
-                          "self.settings.os != 'linux'"]
-            if ("def config_options(self):" in low or "def configure(self):" in low) \
-                    and any(r in low for r in remove_fpic_option):
-                out.success("OK. 'fPIC' option found and apparently well managed")
-            elif "def configure(self):" in low and \
-                 "raise conaninvalidconfiguration" in low and \
-                 any(r in low for r in os_compare):
-                out.success("OK. 'fPIC' option found and apparently this recipe is not " \
-                            "supported for Windows")
-            else:
-                out.error("'fPIC' option not managed correctly. Please remove it for Windows "
-                          "configurations: del self.options.fpic")
+        has_fpic = conanfile.options.get_safe("fPIC") or conanfile.options.get_safe("fpic")
+        if conanfile.settings.get_safe("os") == "Windows" and has_fpic:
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -155,8 +157,17 @@ def test(out):
                     and any(r in low for r in remove_fpic_option):
                 out.success("OK. 'fPIC' option found and apparently well managed")
             else:
-                out.error("'fPIC' option not managed correctly. Please remove it for Windows "
-                          "configurations: del self.options.fpic")
+                conan_instance, _, _ = conan_api.Conan.factory()
```
danimtb: I don't like too much the idea os instantiating a Conan API here, but admit is the most robust option... An alternative could be checking for these lines in the recipe:

```
        if self.settings.os == "Windows":
            raise ConanInvalidConfiguration
```

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -114,3 +135,16 @@ def test_conanfile_installer(self):
         self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)] No 'licenses' folder found in package", output)
         self.assertIn("[DEFAULT PACKAGE LAYOUT (KB-H013)] OK", output)
         self.assertIn("[SHARED ARTIFACTS (KB-H015)] OK", output)
+
+    @unittest.skipIf(platform.system() != "Linux", "Only linux")
+    def test_conanfile_linux_only(self):
+        tools.save('conanfile.py', content=self.conanfile_linux_only)
+        output = self.conan(['create', '.', 'linuxonly/version@conan/test'])
+        self.assertIn("[RECIPE METADATA (KB-H003)] OK", output)
```
danimtb: this check is not related to this PR

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -114,3 +135,16 @@ def test_conanfile_installer(self):
...
+        self.assertIn("[FPIC OPTION (KB-H006)] OK", output)
+        self.assertIn("[FPIC MANAGEMENT (KB-H007)] 'fPIC' has not been removed, but this recipe is "
+                      "not supported by Windows.", output)
+        self.assertIn("[LIBCXX MANAGEMENT (KB-H011)] OK", output)
```
danimtb: same here and the following

hooks/conan-center.py
ericLemanissier: the weakness with this alternative is that sometimes it is `if self.settings.os != "Linux":` (https://github.com/conan-io/conan-center-index/pull/123/files#diff-e89e2904d62404d919e6e127764deb39R22)

danimtb: yes, you are right. Maybe this is the best alternative we have

uilianries: yeah, that's why I preferred using Conan API.

uilianries:
> Too much. It will increase the duration of the hook and also will be another maintenance nightmware. Could this check be moved to a `pre_build` where we have the settings and might check only when `conanfile.settings.os == "Windows"`? WDYT?

Sounds good. As the recipe will fail when exporting, pre_build is a safer place to check it.

uilianries:
I've moved fPIC management to post_source, because pre_build doesn't offer conanfile_path.

SSE4:
@uilianries do you have an update? we need this one to fix ODBC build

danimtb:
@uilianries, I think the check could be easier and done in `pre_build()` with something like:

```
if conanfile.settings.os == "Windows" and conanfile.options.fpic:
   raise ...
```
Maybe we have to use `get_safe()` or get the option in another way, but I think that is the idea @lasote was proposing above. It should be more robust that way instead of checking the conanfile content.

uilianries:
@SSE4 @danimtb I think it's better now! I'm using pre_build to check if is windows and fPIC is declared, no more string parser.

hooks/conan-center.py
```diff
@@ -248,6 +230,20 @@ def _is_removing_libcxx():
                         "Can't detect C++ source files but recipe does not remove 'compiler.libcxx'")
 
 
+@raise_if_error_output
+def pre_build(output, conanfile, **kwargs):
+
+    @run_test("KB-H007", output)
+    def test(out):
+        has_fpic = conanfile.options.get_safe("fPIC") or conanfile.options.get_safe("fpic")
```
ericLemanissier: doesn't `get_safe` return `False` if the option exist and has the value `False` ?

uilianries: >  EDIT: ok, it doesn't !(sorry for the noise)
np! thanks for reviewing!

lasote: The option shouldn't be named "fpic". Maybe that check is a `pre-export` one?

danimtb: I agree, but that is another check that should fall into the syntax side. Let's try to get this PR done and go for that later

lasote: So, remove the "or" here

danimtb:
Merging this as Appveyor CI already passed in previous commits (not affected by the latest ones)

PR #90, opened 2019-09-30
Remove cppstd for C projects

### LINKED ISSUE #89: [conan center] check for self.settings.compiler.cppstd
setting `self.settings.compiler.cppstd` should be removed, same check as `self.settings.compiler.libcxx`

uilianries:
- Both libcxx and cppstd should be removed when is a recipe for C project only.

closes #89 
Wiki:

#### **<a name="KB-H022">#KB-H022</a>: "CPPSTD MANAGEMENT"**

If the library is detected as a pure C library (sources doesn't conatain any file with the following extensions: ["cc", "cpp", "cxx", "c++m", "cppm", "cxxm", "h++", "hh", "hxx", "hpp"]) then it has to remove the `compiler.cppstd` subsetting, because the cpp standard library shouldn't affect the binary ID:

```python
class SomeRecipe(ConanFile):
    ...

    def configure(self):
        del self.settings.compiler.cppstd
```

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -228,37 +228,31 @@ def pre_source(output, conanfile, conanfile_path, **kwargs):
 @raise_if_error_output
 def post_source(output, conanfile, conanfile_path, **kwargs):
 
-    def _is_removing_libcxx():
-        conanfile_content = tools.load(conanfile_path)
-        low = conanfile_content.lower()
-        conf = "def configure(self):"
-        conf2 = "del self.settings.compiler.libcxx"
-        return conf in low and conf2 in low
-
-    @run_test("KB-H011", output)
-    def test(out):
+    def _is_pure_c():
         if not _is_recipe_header_only(conanfile):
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -226,25 +227,39 @@ def test(out):
...
+    def test(out):
+
+        def _is_removing_cppstd():
+            conanfile_content = tools.load(conanfile_path)
+            low = conanfile_content.lower()
+            conf = "def configure(self):"
+            conf2 = "del self.settings.compiler.cppstd"
+            return conf in low and conf2 in low
+
+        if _is_removing_libcxx() and not _is_removing_cppstd():
```
danimtb: I would not make this conditional to libcxx but to a pure C function check that could be used both for detection in KB-H011 and here (KB-H022)

lasote: Make them independent as @danimtb suggested. You can "refactor" the detection of "C/CXX" extensions.

PR #94, opened 2019-09-30
Validate shebang

### LINKED ISSUE #84: [conan center] check meta lines
check the following meta lines are not in `conanfile.py` and `test_package/conanfile.py`:

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
# vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
```

- shebang
- vim
- coding header

uilianries:
Validate shebang and other meta data in Conan recipes.

closes #84

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -25,7 +25,8 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H018": "LIBTOOL FILES PRESENCE",
              "KB-H019": "CMAKE FILE NOT IN BUILD FOLDERS",
              "KB-H020": "PC-FILES",
-             "KB-H021": "MS RUNTIME FILES"}
+             "KB-H021": "MS RUNTIME FILES",
+             "KB-H025": "META LINES"}
 
 
 class _HooksOutputErrorCollector(object):
@@ -188,6 +189,17 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
         if total_size_kb > max_folder_size:
             out.error("The size of your recipe folder ({} KB) is larger than the maximum allowed"
                       " size ({}KB).".format(total_size_kb, max_folder_size))
+
+    @run_test("KB-H025", output)
+    def test(out):
+        if "# -*- coding: utf-8 -*-" in conanfile_content:
+            out.error("PEP 263 (encoding) is not allowed in the conanfile. Remove `coding: utf-8`")
+        if "#!/usr/bin/env python" in conanfile_content or \
+           "#!/usr/local/bin/python" in conanfile_content or \
+           "#!/usr/bin/python" in conanfile_content:
+            out.error("Shebang detected in your recipe. Remove the Python version from the recipe")
+        if "# vim:" in conanfile_content:
+            out.error("vim editor configuration detected in your recipe. Remove it from the recipe")
 
 
 @raise_if_error_output
--- a/tests/test_hooks/conan-center/test_conan-center.py
+++ b/tests/test_hooks/conan-center/test_conan-center.py
@@ -67,6 +67,7 @@ class ConanCenterTests(ConanClientTestCase):
         self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)] No 'licenses' folder found in package", output)
         self.assertIn("[DEFAULT PACKAGE LAYOUT (KB-H013)] OK", output)
         self.assertIn("[SHARED ARTIFACTS (KB-H015)] OK", output)
+        self.assertIn("[META LINES (KB-H025)] OK", output)
 
     def test_conanfile_header_only(self):
         tools.save('conanfile.py', content=self.conanfile_header_only)
@@ -83,6 +84,7 @@ class ConanCenterTests(ConanClientTestCase):
         self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)] No 'licenses' folder found in package", output)
         self.assertIn("[DEFAULT PACKAGE LAYOUT (KB-H013)] OK", output)
         self.assertIn("[SHARED ARTIFACTS (KB-H015)] OK", output)
+        self.assertIn("[META LINES (KB-H025)] OK", output)
 
     def test_conanfile_header_only_with_settings(self):
         tools.save('conanfile.py', content=self.conanfile_header_only_with_settings)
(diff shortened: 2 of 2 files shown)


hooks/conan-center.py
```diff
@@ -189,6 +190,17 @@ def test(out):
             out.error("The size of your recipe folder ({} KB) is larger than the maximum allowed"
                       " size ({}KB).".format(total_size_kb, max_folder_size))
 
+    @run_test("KB-H025", output)
+    def test(out):
+        if "# -*- coding: utf-8 -*-" in conanfile_content:
+            out.error("PEP 263 (encoding) is not allowed in the conanfile. Remove `coding: utf-8`")
+        if "#!/usr/bin/env python" in conanfile_content or \
```
SSE4: I think just check for `#!` at the beginning should be enough, as shebang may have various forms, depends on the system

hooks/conan-center.py
```diff
@@ -189,6 +190,17 @@ def test(out):
             out.error("The size of your recipe folder ({} KB) is larger than the maximum allowed"
                       " size ({}KB).".format(total_size_kb, max_folder_size))
 
+    @run_test("KB-H025", output)
+    def test(out):
+        if "# -*- coding: utf-8 -*-" in conanfile_content:
```
SSE4: just in case, check for `# -*- coding:`, without `utf-8`, as there are might be other encodings

hooks/conan-center.py
SSE4: and check that it's really a beginning of a the file, because the following is legal:
```
tools.save("#!/usr/bin/env python")
```

uilianries: done!

hooks/conan-center.py
uilianries: @SSE4 Done!

uilianries:
> Attend the @SSE4 comments, add test to validate a shebang in the middle is not affecting and write the KB description (that will go to the wiki) on the PR description to help to understand.

@lasote I've added a new entry on Wiki: https://github.com/conan-io/conan-center-index/wiki/Error-Knowledge-Base#kb-h025-meta-lines

hooks/conan-center.py
```diff
@@ -191,6 +192,19 @@ def test(out):
             out.error("The size of your recipe folder ({} KB) is larger than the maximum allowed"
                       " size ({}KB).".format(total_size_kb, max_folder_size))
 
+    @run_test("KB-H025", output)
+    def test(out):
+        lines = conanfile_content.split('\n')
```
danimtb:
```suggestion
        lines = conanfile_content.splitlines()
```

hooks/conan-center.py
```diff
@@ -191,6 +192,19 @@ def test(out):
...
+        lines_range = 5 if len(lines) > 5 else len(lines)
```
danimtb: Not sure if these are ALWAYS at the top. I am thinking about people that write header comments or license things, that could be before the encoding and shebang lines. But I am not sure and probably checking the 5 firest lines is safe enough

hooks/conan-center.py
```diff
@@ -191,6 +192,19 @@ def test(out):
...
+        for index in range(0,lines_range):
+            if "# -*- coding:" in lines[index] or \
+               "# coding=" in lines[index]:
+                out.error("PEP 263 (encoding) is not allowed in the conanfile. Remove `coding: utf-8`")
+            if "#!" in lines[index]:
+                out.error("Shebang (#!) detected in your recipe. Remove it from the recipe")
+            if "# vim:" in lines[index]:
+                out.error("vim editor configuration detected in your recipe. Remove it from the recipe")
```
danimtb: Could we print the offending line and maybe the number?

hooks/conan-center.py
uilianries: I agree. Yesterday I saw a recipe which has the vim config at the footer. As we are only checking the conanfile.py I think validating the first 5 lines and last 3 lines is good enough.

hooks/conan-center.py
uilianries: excellent idea! I gonna improve it.

uilianries:
#### **<a name="KB-H025">#KB-H025</a>: "META LINES"**

Some specific metadata are not allowed in Conan recipes:

- Shebang) to specify Python version:

```python
#!/usr/bin/env python  # not allowed
#!/usr/local/bin/python  # not allowed

class SomeRecipe(ConanFile):
    ...
```

- [Vim](https://www.vim.org/) configuration:

```python
# vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4 # not allowed

class SomeRecipe(ConanFile):
    ...
```

- Encoding:

```python
# -*- coding: utf-8 -*-  # not allowed
# coding=utf-16          # not allowed

class SomeRecipe(ConanFile):
    ...
```

hooks/conan-center.py
```diff
@@ -191,6 +192,29 @@ def test(out):
...
 
+    @run_test("KB-H025", output)
+    def test(out):
+        def _search_for_metaline(from_line, to_line, lines):
+            for index in range(from_line,to_line):
+                line_number = index + 1
+                if "# -*- coding:" in lines[index] or \
+                   "# coding=" in lines[index]:
+                    out.error("PEP 263 (encoding) is not allowed in the conanfile. " \
+                              "Remove `coding: utf-8` from the line {}".format(line_number))
```
lasote: We should say `"Remove the line containing coding: utf-8`, right? Not only the `coding: utf-8`

uilianries:
all conflicts fixed

danimtb:
Some tests already passed so I am merging this as it is purely related to recipe syntax

jgsogo:
**encoding**: I agree it is not needed for Python3 (then I would validate-force that the file is utf-8 readable), but Conan still supports Python2 and declaring the encoding could be needed if you want to write unicode strings, i.e.: the name of the author could appear in the description of the recipe (or comments). This is something that has already happened to us in the past with the CTRE recipe contributed by Hana herself.

Also, I'm not sure if going against a PEP is something we should think twice.

If we keep this validation, then we should check all the valid candidates, use regex `^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)` (only need to check first or second line).

lasote:
Summary, I like the code without "Ñ" I've never use it, I don't want to unnecessarily start fighting with encodings.

PR #112, opened 2019-10-17
#100 do not allow system requirements

### LINKED ISSUE #100: [conan center] check there is no system_requirements
I think we have previously agreed to disallow `system_requirements` in conan-center-index
so, it would be nice to check in hook there are no `system_requirements` methods
all exceptions will be white-listed in hook I guess

uilianries:
As discussed before, `system_requirements` is evil and should be avoided. This hook looks its declaration in the conanfile.

closes #100 

Wiki:

#### **<a name="KB-H032">#KB-H032</a>: "SYSTEM REQUIREMENTS"**

System requires can be used as an option when a Conan package is not available the same package can be installer system package manager. However, it can cause  reproducibility problems, since the package may vary according the distribution or OS. Conan is not able to track its metadata, so that, installing system packages by recipe is not allowed.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -270,8 +270,13 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H031", output)
     def test(out):
-        if "def system_requirements" in conanfile_content:
-            out.error("The method 'system_requirements' is not allowed in the recipe.")
+        if "def system_requirements" in conanfile_content and \
+           "SystemPackageTool" in conanfile_content:
+            import re
+            match = re.search(r'(\S+)\s?=\s?SystemPackageTool', conanfile_content)
+            if ("SystemPackageTool().install" in conanfile_content) or \
+               (match and "{}.install".format(match.group(1)) in conanfile_content):
+                out.error("The method 'SystemPackageTool.install' is not allowed in the recipe.")
 
 
 @raise_if_error_output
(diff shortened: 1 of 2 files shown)


memsharded:
Just be aware that there might be a couple of them like OpenCV that would be extremely challenging to get Gtk-dev components in the system correctly. 
Maybe we should block just the install, but not the check of system dependencies.

uilianries:
> Just be aware that there might be a couple of them like OpenCV that would be extremely challenging to get Gtk-dev components in the system correctly.
> Maybe we should block just the install, but not the check of system dependencies.

I see. So should we add a whitelist, e.g OpenCV

memsharded:
No, the OpenCV recipe must check that those system packages are installed or fail otherwise listing the packages to install. Do not allow to install system packages by recipes, but check them.

uilianries:
understood, thanks for clarifying

uilianries:
All conflicts has been solved.

SSE4:
@uilianries conflicts again

uilianries:
All conflicts has been solved. (2)

Croydon:
Could we move this one forward?

danimtb:
I think we should be prepared to whitelist some of the packages as we are planning to introduce some system requirements as "proxy recipes". @uilianries could you please include that in this PR and whitelist `libusb` at the moment? Thanks!!

uilianries:
@danimtb Done, libusb is part of allowlist now.

PR #113, opened 2019-10-18
Sse cpp_info.frameworks instead of cpp_info.exe_link_flags

### LINKED ISSUE #108: [conan center] advice to use cpp_info.frameworks instead of cpp_info.exe_link_flags
some old recipes use code like:
```
self.cpp_info.shared_link_flags.append("-framework CoreAudio")
```
since conan 1.19, it's better to use just:
```
self.cpp_info.frameworks.append("CoreAudio")
```
the hook may advice the new systax over the old one

uilianries:
Based on [Apple Framework support](https://docs.conan.io/en/latest/howtos/link_apple_framework.html), now all recipes using `exe_link_flags` for OSX framework should use `cpp_info.frameworks` instead.

closes #108

Wiki: 

#### **<a name="KB-H032">#KB-H032</a>: "APPLE FRAMEWORK"**

When [liking Apple frameworks](https://docs.conan.io/en/latest/howtos/link_apple_framework.html#how-to-link-with-apple-frameworks) into your current recipe, instead of `self.cpp_info.shared_link_flags`, use `self.cpp_info.frameworks`:

```python
    def package_info(self):
        self.cpp_info.frameworks.append("CoreAudio")
```

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -32,7 +32,9 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H025": "META LINES",
              "KB-H027": "CONAN CENTER INDEX URL",
              "KB-H028": "CMAKE MINIMUM VERSION",
-             "KB-H029": "TEST PACKAGE - RUN ENVIRONMENT"}
+             "KB-H029": "TEST PACKAGE - RUN ENVIRONMENT",
+             "KB-H032": "APPLE FRAMEWORK"
+        }
 
 
 class _HooksOutputErrorCollector(object):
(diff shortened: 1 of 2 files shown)


uilianries:
All conflicts has been solved.

hooks/conan-center.py
```diff
@@ -276,6 +277,11 @@ def test(out):
             out.error("The 'RunEnvironment()' build helper is no longer needed. "
                       "It has been integrated into the self.run(..., run_environment=True)")
 
+    @run_test("KB-H033", output)
+    def test(out):
+        if "cpp_info.shared_link_flags" in conanfile_content and "-framework" in conanfile_content:
```
SSE4: that might trigger false positives, need to reduce scope of test

uilianries: can you elaborate some idea?

SSE4: having something like:
```
self.cpp_info.shared_link_flags = ["-B,Symbolic"]
...
tools.replace_in_file("Makefile", "-framework Foundation", "-framework CoreFoundation")
```
will result in false positive I believe

PR #121, opened 2019-10-30
#102 Accept script files

### LINKED ISSUE #102: [bug] Conan Center hook is not able to deal with shell script recipe
https://github.com/conan-io/conan-center-index/pull/156#issuecomment-536999606

uilianries:
This won't fix #102 totally, but it helps. As we can't deduce if a package will install headers, the current rule for _is_header_only will check only few characteristics in the recipe and run during pre-export. We would need to valid whether package has installed header files in include folder. 

The CCCL project, for instance. According the settings and package id it is header-only, it doesn't export any header, only a bash script on bin folder.

fixes #102

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,17 +27,17 @@ jobs:
       dist: xenial
       sudo: true
 
+    - stage: Linux - Conan 1.19
+      name: Python 3.7
+      python: 3.7
+      env: TOXENV=py37-conan119
+      dist: xenial
+      sudo: true
+
     - stage: Linux - Conan 1.18
       name: Python 3.7
       python: 3.7
       env: TOXENV=py37-conan118
-      dist: xenial
-      sudo: true
-
-    - stage: Linux - Conan 1.17
-      name: Python 3.7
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -27,17 +27,17 @@ jobs:
       dist: xenial
       sudo: true
 
-    - stage: Linux - Conan 1.18
+    - stage: Linux - Conan 1.19
```
danimtb: please remove changes in this file

tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
     conandev: https://github.com/conan-io/conan/archive/develop.tar.gz
     conancurrent: conan
     conan118: conan>=1.18,<1.19
-    conan117: conan>=1.17,<1.18
```
danimtb: please remove changes in this file

.travis.yml
uilianries: done

tox.ini
uilianries: done

PR #124, opened 2019-11-01
#123 Custom attributes must be protected

### LINKED ISSUE #123: Custom attributes must be protected
As discussed in the past, only ConanFile attributes (e.g. topics, options, settings, ...) should be public. Custom attributes (e.g. _build_subfolder, _source_subfolder) must be declared as protect to avoid any possible future conflict.

uilianries:
- I need to compare if the attributes in the conanfile passed to this hook, contains more attributes and if they are protected. However, I need to compare using a default ConanFile object, which is not loaded by `basic_load`, because it requires a conan recipe (file) to be loaded. Also, the loaded injects some attributes, which are not present in the Conanfile, so I checked them adding the attributes by their names (conan_data and python_requires)

closes #123

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -7,7 +7,7 @@ from collections import defaultdict
 import yaml
 from logging import WARNING, ERROR, INFO, DEBUG, NOTSET
 
-from conans import tools, Settings
+from conans import tools, Settings, ConanFile
 
 kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H002": "REFERENCE LOWERCASE",
@@ -37,7 +37,9 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H028": "CMAKE MINIMUM VERSION",
              "KB-H029": "TEST PACKAGE - RUN ENVIRONMENT",
              "KB-H030": "CONANDATA.YML FORMAT",
-             "KB-H031": "CONANDATA.YML REDUCE"}
+             "KB-H031": "CONANDATA.YML REDUCE",
+             "KB-H034": "CUSTOM ATTRIBUTES",
+            }
 
 
 class _HooksOutputErrorCollector(object):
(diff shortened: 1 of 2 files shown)


jgsogo:
A little bit afraid of this one. It can break if we add to Conan any field (also dynamic ones) and we might not notice it until _all_ the PRs in ConanCenter fail. For example, here you are missing `python_requires_extend`, but there are other fields we all might not be aware right now.

This is a nice hook, but I will make it a warning, not an error. wdyt?

uilianries:
> It can break if we add to Conan any field (also dynamic ones) and we might not notice it until all the PRs in ConanCenter fail.

Do you mean protected Conan fields? But we are not allowed to use it in CCI because we can't ensure them for breaking, we only use public and exported methods.

> For example, here you are missing python_requires_extend, but there are other fields we all might not be aware right now.

I didn't understand, `python_requires_extend` is public, it won't be blocked by this hook. Is there any protected field that we are using and should be public?

> This is a nice hook, but I will make it a warning, not an error. wdyt?

The problem using warning level is that we need to build when reviewing and checking the log, OR read output logs from CI. IMO it doesn't work, repetitive human review is susceptible to failure.

SSE4:
`python_requires_extend` is mentioned twice in the docs, but appears to be undocumented on its own. what does it do?

uilianries:
@SSE4 Indeed the example on https://docs.conan.io/en/latest/extending/python_requires.html#extending-base-classes shows its usage, but I didn't find any reference with details.

jgsogo:
My concern is that you have added `conan_data` and `python_requires` by hand. Any other field can pop up in the future and this hook will fail. Probably this is not as bad as I'm thinking about because we can iterate really fast in this repo to handle the new member if needed. So ok, not a problem.

@SSE4 `python_requires_extend` is the way to inherit from a base class inside a python requires. Conan injects this class as a base class to the conanfile in runtime. Why? We need to be able to load the python file with the recipe and the python requires information before resolving the classes/methods from `python_requires` (if we use Python inheritance we need to retrieve `python_requires` first and the interpreter will use the base class). This let Conan control the `python_requires`.

jgsogo (dismissed):

uilianries:
Nice explanation, thanks @jgsogo !

SSE4:
@jgsogo thanks for the explanation. I believe `python_requires_extend` should be documented [there](https://docs.conan.io/en/latest/reference/conanfile/attributes.html). I'll create an issue for docs.

SSE4 (dismissed):

hooks/conan-center.py
```diff
@@ -367,6 +368,19 @@ def test(out):
         if "def imports" in test_package_conanfile:
             out.error("The method `imports` is not allowed in test_package/conanfile.py")
 
+    @run_test("KB-H035", output)
+    def test(out):
+        mock = ConanFile(conanfile.output, None)
+        valid_attrs = [attr for attr in dir(mock) if not callable(attr)] + ['conan_data', 'python_requires']
```
jgsogo, line 374: At least, here we need `python_requires_extend` too. And for `1.26` we should handle new attributes/functions that could result of this PR https://github.com/conan-io/conan/pull/6945

PR #127, opened 2019-11-01
Custom methods must be protected

### LINKED ISSUE #126: Custom method must be protected
Same as #123, but for methods only

take a look on the notes here: https://docs.conan.io/en/latest/reference/conanfile.html#conanfile-py

uilianries:
- Any method declared as public, must be provided from ConanFile
- Any custom method must be declared as protected
- Any private method is not be accepted

closes #126 

Wiki:

#### **<a name="KB-H036">#KB-H036</a>: "CUSTOM METHODS"**

When a new method have to be added to the [recipe](https://docs.conan.io/en/latest/reference/conanfile.html#conanfile-py), and it is not a default one provided by ConanFile, it must added as protected:

```python
class Recipe(ConanFile):
    def configure(self): # default methods are public
        pass

    def _configure_cmake(self): # custom methods are protected
        pass

    def __private_routine(self): # private should not be used. Use protected 
instead.
	     pass

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -7,7 +7,7 @@ from collections import defaultdict
 import yaml
 from logging import WARNING, ERROR, INFO, DEBUG, NOTSET
 
-from conans import tools, Settings
+from conans import tools, Settings, ConanFile
 
 kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H002": "REFERENCE LOWERCASE",
@@ -37,7 +37,9 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H028": "CMAKE MINIMUM VERSION",
              "KB-H029": "TEST PACKAGE - RUN ENVIRONMENT",
              "KB-H030": "CONANDATA.YML FORMAT",
-             "KB-H031": "CONANDATA.YML REDUCE"}
+             "KB-H031": "CONANDATA.YML REDUCE",
+             "KB-H036": "CUSTOM METHODS",
+        }
 
 
 class _HooksOutputErrorCollector(object):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: hooks/conan-center.py [hunk 1]
```diff
@@ -319,6 +321,31 @@ def _not_allowed_entries(info, allowed_entries):
...
+                    if callable(getattr(conanfile, member)):
+                        methods.append(str(member))
+                except:
+                    methods.append(str(member))
+            return methods
+
+        mock = ConanFile(conanfile.output, None)
+        valid_attrs = get_methods(mock)
+        current_attrs = get_methods(conanfile)
+        invalid_attrs = re.findall(r"def (__.*[^_])\s?\(", conanfile_content, re.MULTILINE)
```
SSE4: I don't get it, you query valid attrs and current attrs, why query invalid attrs with different method?
isn't result the same as `current_attrs - valid_attrs`?

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -411,3 +414,38 @@ def test_cmake_minimum_version(self):
...
+            def __abs__(self):
+                return self.version
+
+            def _baz(self):
+                self.output.info("qux")
+
+        """)
+        tools.save('conanfile.py', content=conanfile)
+        output = self.conan(['create', '.', 'name/version@user/test'])
+        self.assertIn("ERROR: [CUSTOM METHODS (KB-H036)] Custom methods must be declared as "
```
SSE4: might be too strict, as we may define some magic methods (`__str__`, `__eq__`,  `__new__`, `__init__`, `__enter__`, `__dir__`, etc)?

hooks/conan-center.py
uilianries: private methods are not filtered by `dir` method. e.g. `def __foo(self):`

tests/test_hooks/conan-center/test_conan-center.py
uilianries: but it's the exception you know. It sounds like cccl and other few recipes. We know that is not a regular case, but new contributors can use private instead of protected when writing a recipe, for instance.

SSE4:
@uilianries also, I wonder, why do we forbid to use private members? is it somehow bad/harmful?

uilianries:
@SSE4 private attributes are not bad, but how many times we needed them for a recipe? Maybe twice all this time, only for workarounds and helpers, which are not the regular flow.

PR #131, opened 2019-11-07
Validate non-ascii characters 

### LINKED ISSUE #130: Warning non ascii recipes
As you know, Python 2 loads the python files as ascii by default. As we removed the shebang from our recipes, we need to check manually or wait for someone open an issue complaining about the recipe support.

@madebr has a good idea, add a hook which tries to load the conanfile with ascii encoding. As Python 2 is not a requirement, we can show a warning message instead of an error.

uilianries:
As Python 2 is close to its retirement, so I preferred keep this hook as an error. After 2020 we can change it to warning instead.

Related issue: https://github.com/conan-io/conan-center-index/pull/278#discussion_r343837509

/cc @madebr @SSE4 

closes #130

Wiki:

#### **<a name="KB-H038">#KB-H038</a>: "ASCII SUPPORT"**

Characters non-ascii are not compatible with Python 2, so they should be avoided.

### DIFF
--- a/tests/test_hooks/conan-center/test_conan-center.py
+++ b/tests/test_hooks/conan-center/test_conan-center.py
@@ -465,15 +465,14 @@ class ConanCenterTests(ConanClientTestCase):
                       'Remove \'author = (\'foo\', \'bar\')', output)
 
     def test_invalid_ascii_recipe(self):
-        conanfile = textwrap.dedent("""\
-        from conans import ConanFile
-        class AConan(ConanFile):
-            description = "open-source file compression that uses the Burrows–Wheeler algorithm."
-
-            def configure(self):
-                pass
-        """)
+        conanfile = textwrap.dedent(b"""
+            from conans import ConanFile
+            class AConan(ConanFile):
+                description = "file compression that uses the Burrows\xe2\x80\x93Wheeler algorithm"
+                def configure(self):
+                    pass
+        """.decode(encoding="utf8"))
         tools.save('conanfile.py', content=conanfile.replace("{}", ""))
         output = self.conan(['create', '.', 'name/version@user/test'])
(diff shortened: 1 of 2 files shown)


tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -462,3 +463,17 @@ def configure(self):
         output = self.conan(['create', '.', 'name/version@user/test'])
         self.assertIn('ERROR: [NO AUTHOR (KB-H037)] Conanfile should not contain author. '
                       'Remove \'author = (\'foo\', \'bar\')', output)
+
+    def test_invalid_ascii_recipe(self):
+        conanfile = textwrap.dedent("""\
+        from conans import ConanFile
+        class AConan(ConanFile):
+            description = "open-source file compression that uses the Burrows–Wheeler algorithm."
```
madebr: By including this character, the `test_conan-center.py` source itself becomes non-ascii.
Maybe create the conanfile source as follows?
```
conanfile = b"""

        from conans import ConanFile
        class AConan(ConanFile):
            description = "open-source file compression that uses the Burrows\xe2\x80\x93Wheeler algorithm."
            def configure(self):
                pas
"""
```

uilianries: hahaha! It's true! I gonna do it.

uilianries: this was funny! it's fixed now!

madebr:
I think this hook is ok in its current state.
Can this be merged?

danimtb:
let's update it and merge it

uilianries:
@danimtb All green again!

danimtb:
I totally missed this one and I think it is a duplicate of #199

uilianries:
And I created both PR :facepalm:

PR #138, opened 2019-11-28
#135 Revision should not be added to the recipe

### LINKED ISSUE #135: [conan center] check there are no scm and revision mode attributes

uilianries:
Both `scm` and `revision_mode` should not used in the recipes

closes #135 

WIKI:

#### **<a name="KB-H039">#KB-H039</a>: "NOT ALLOWED ATTRIBUTES"**

Some attributes are experimental features or may influence the package id, so that they should be avoided.


danimtb:
I think the message "NO REVISION" is a bit unclear to me. I would make the message more generic like "NOT ALLOWED ATTRIBUTES" (we need a better name) and indicate the attribute that is not allowed.

hooks/conan-center.py
```diff
@@ -345,6 +346,18 @@ def test(out):
...
+    @run_test("KB-H039", output)
+    def test(out):
+        forbidden_attrs = []
+        if getattr(conanfile, "scm", None):
+            forbidden_attrs.append("scm")
+        if re.search(r"revision_mode\s=", conanfile_content):
+            forbidden_attrs.append("revision_mode")
+
+        if forbidden_attrs:
+            out.error("Conanfile should not contain attributes related to revision. "
```
danimtb: please change the error message

hooks/conan-center.py
```diff
@@ -345,6 +346,18 @@ def test(out):
...
+    @run_test("KB-H039", output)
+    def test(out):
+        forbidden_attrs = []
+        if getattr(conanfile, "scm", None):
+            forbidden_attrs.append("scm")
+        if re.search(r"revision_mode\s=", conanfile_content):
+            forbidden_attrs.append("revision_mode")
+
+        if forbidden_attrs:
+            out.error("Conanfile should not contain the follow attributes: '{}'. Remove them."
```
danimtb:
```suggestion
            out.error("Conanfile should not contain attributes: '{}'"
```

uilianries: @danimtb Done!

danimtb (approved):

SSE4 (approved):

PR #142, opened 2020-01-07
Validate cpp_info.name

uilianries:
- Update Conan version for testing
- Add new hook which checks cpp_info.name is declared
- Add tests for hook KB-H040

Related issue: https://github.com/conan-io/conan/issues/6269#issuecomment-570195413

Closes https://github.com/conan-io/conan/issues/6001

/cc @Morwenn

Wiki:

#### **<a name="KB-H040">#KB-H032</a>: "NO TARGET NAME"**

According the Conan issue [#6269](https://github.com/conan-io/conan/issues/6269), the attribute `cpp_info.name` should be avoided for Conan Center Index in favor of `cpp_info.names["cmake_find_package"]` and `cpp_info.names["cmake_find_multi_package"]`.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -364,14 +364,13 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
     def test(out):
         if "self.cpp_info.name =" in conanfile_content:
             out.error("CCI uses the name of the package for cmake generator."
-                      " Conanfile should not contain 'self.cpp_info.name'."
                       " Use 'cpp_info.names' instead.")
 
         for generator in ["cmake", "cmake_multi"]:
             if "self.cpp_info.names['{}']".format(generator) in conanfile_content or \
                'self.cpp_info.names["{}"]'.format(generator) in conanfile_content:
-                out.error("CCI uses the name of the package for cmake generator. "
-                          "Conanfile should not contain 'self.cpp_info.names['{}']'. "
+                out.error("CCI uses the name of the package for {0} generator. "
+                          "Conanfile should not contain 'self.cpp_info.names['{0}']'. "
                           " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.".format(generator))
 
 @raise_if_error_output
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -39,6 +39,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H029": "TEST PACKAGE - RUN ENVIRONMENT",
              "KB-H030": "CONANDATA.YML FORMAT",
              "KB-H031": "CONANDATA.YML REDUCE",
+             "KB-H032": "SYSTEM REQUIREMENTS",
              "KB-H034": "TEST PACKAGE - NO IMPORTS()",
              "KB-H037": "NO AUTHOR",
              "KB-H040": "NO TARGET NAME",
@@ -282,6 +283,19 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
             out.error("The 'RunEnvironment()' build helper is no longer needed. "
                       "It has been integrated into the self.run(..., run_environment=True)")
 
+    @run_test("KB-H032", output)
+    def test(out):
+        if conanfile.name in ["libusb"]:
+            out.info("'{}' is part of the allowlist.".format(conanfile.name))
+            return
+        if "def system_requirements" in conanfile_content and \
+           "SystemPackageTool" in conanfile_content:
+            import re
+            match = re.search(r'(\S+)\s?=\s?SystemPackageTool', conanfile_content)
+            if ("SystemPackageTool().install" in conanfile_content) or \
+               (match and "{}.install".format(match.group(1)) in conanfile_content):
+                out.error("The method 'SystemPackageTool.install' is not allowed in the recipe.")
+
     @run_test("KB-H030", output)
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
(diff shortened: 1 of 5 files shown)


danimtb:
Please add the description of the error KB in the description of this PR

danimtb:
@jgsogo there is a nightly job doing just `conan export` over all the recipes with hooks activated in order to test them. I think that should be enough, right?

jgsogo:
Yes, that one is enough. The idea is to check if using `cpp_info.name` is the way to go for some recipe and then we cannot activate this hook worlwide.

danimtb:
I see some packages that are still using cpp_info.name: bzip2, entt and openssl. I would like to have those updated first before merging this

uilianries:
> I see some packages that are still using cpp_info.name: bzip2, entt and openssl. I would like to have those updated first before merging this

OpenSSL: https://github.com/conan-io/conan-center-index/pull/604
EnTT: https://github.com/conan-io/conan-center-index/pull/673
Bzip2: https://github.com/conan-io/conan-center-index/pull/672

danimtb:
@uilianries https://github.com/conan-io/hooks/pull/142#issuecomment-571943400

danimtb:
Please update this PR when possible and add the description of the KB in the description of this PR

uilianries:
@danimtb Conflicts has been fixed. Also, I've added the wiki description, please take a look. Thanks for pinging me.

hooks/conan-center.py
```diff
@@ -359,6 +360,16 @@ def test(out):
...
+    @run_test("KB-H040", output)
+    def test(out):
+        if "self.cpp_info.name =" in conanfile_content:
+            out.error("Conanfile should not contain 'self.cpp_info.name'."
+                      " Use 'cpp_info.names' instead.")
+
+        if "self.cpp_info.names['cmake']" in conanfile_content or \
+           'self.cpp_info.names["cmake"]' in conanfile_content:
+            out.error("Conanfile should not contain 'self.cpp_info.names['cmake']'."
+                      " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.")
```
jgsogo: I think that the reason to avoid `self.cpp_info.names['cmake']` is because it should match the name of the package

jgsogo:
> According the Conan issue [#6269](https://github.com/conan-io/conan/issues/6269), the attribute `cpp_info.name` has been deprecated in favor of `cpp_info.names["cmake_find_package"]` and `cpp_info.names["cmake_find_multi_package"]`.

I wouldn't say that `cpp_info.name` is deprecated, just say that we prefer not to use it in Conan Center 😉

hooks/conan-center.py
uilianries: okay, but what message should we use instead?

uilianries:
> > According the Conan issue [#6269](https://github.com/conan-io/conan/issues/6269), the attribute `cpp_info.name` has been deprecated in favor of `cpp_info.names["cmake_find_package"]` and `cpp_info.names["cmake_find_multi_package"]`.
> 
> I wouldn't say that `cpp_info.name` is deprecated, just say that we prefer not to use it in Conan Center wink

I just updated it now, thanks for reviewing

hooks/conan-center.py
jgsogo: Something like: _"Conan Center uses the name of the package for `cmake` generator, do not change it here"_, nevertheless, if this is needed, would it be possible to add the package to a whitelist or something similar?

Morwenn:
As mentioned in conan-io/conan-center-index#787 the restriction that applies to `cmake` should probably apply to `cmake_multi` too for consistency.

hooks/conan-center.py
uilianries: Depending the package, is possible add it to a allow list. However, that hook will prevent unsuspecting people.

uilianries:
@jgsogo @Morwenn I just update the branch, please, review again.

jgsogo:
@Morwenn, I think that everything that applies to 'cmake' should apply to 'cmake_multi' too (at least in 'cpp_info'). And the same for 'cmake_find_package' and 'visual_studio', and any other '_multi' out there. Probably someone should open the issue :D

hooks/conan-center.py
```diff
@@ -359,6 +360,19 @@ def test(out):
                 author = '"%s"' % author
             out.error("Conanfile should not contain author. Remove 'author = {}'".format(author))
 
+    @run_test("KB-H040", output)
+    def test(out):
+        if "self.cpp_info.name =" in conanfile_content:
+            out.error("CCI uses the name of the package for cmake generator."
```
Morwenn: Won't the first line confuse users who didn't try to use `cpp_info.name` for the CMake generators?

hooks/conan-center.py
```diff
@@ -359,6 +360,19 @@ def test(out):
...
+                      " Conanfile should not contain 'self.cpp_info.name'."
+                      " Use 'cpp_info.names' instead.")
+
+        for generator in ["cmake", "cmake_multi"]:
+            if "self.cpp_info.names['{}']".format(generator) in conanfile_content or \
+               'self.cpp_info.names["{}"]'.format(generator) in conanfile_content:
+                out.error("CCI uses the name of the package for cmake generator. "
```
Morwenn: Maybe replace `cmake` in this string with the value of `generator`?

hooks/conan-center.py
uilianries: done, updated

hooks/conan-center.py
uilianries: done, updated

Morwenn:
@uilianries It's good for me, thanks for everything 😄 

With that in it's probably the better answer we can give to my original issue conan-io/conan#6001 right now. It will probably be good enough to close the issue since I don't believe there is a better solution to the problem right now.

uilianries:
@Morwenn thanks a lot for reviewing!

PR #143, opened 2020-01-09
Add os to KB-H030

uilianries:
The idea here comes from NodeJS recipe, where we are using one different installer per OS:

https://github.com/bincrafters/conan-nodejs_installer/blob/testing/10.15.0/conanfile.py#L23

We could manage is directly in Conanfile.py, but it would look terrible. Allowing OS into KB-H030 we can classify the correct installer based on OS;


uilianries:
Forget about, I found a better way

PR #153, opened 2020-02-05
Testing using py35 and py38

jgsogo:
I've many doubts about the versions we should use in this CI, what do you think?
 * Windows users will likely install the latest Python available, so they will be using py38 or py37 depending on when they did install their system
 * Unix users might use the one installed in the system, and it could be py3.5 😕 

wdyt? Do we care about old versions? Just test with the _latest_? Only two _latest_ versions? Try to resemble the ecosystem out there?

Meanwhile, I want to know if these hooks works with py38 or there is any issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,10 @@ jobs:
       python: 3.7
       env: TOXENV=py37-conandev
       dist: xenial
-      sudo: true
+    - name: Python 3.8
+      python: 3.8
+      env: TOXENV=py38-conandev
+      dist: xenial
 
     # All Linux first, check versions
     - stage: Linux - Conan Current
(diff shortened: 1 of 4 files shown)


uilianries:
> Unix users might use the one installed in the system, and it could be py3.5 confused

Hahaha define "system". I run Arch, my Python version is 3.8.1. Well, that's the problem of linux, many distributions.

On Windows, I'm not sure, but I use Chocolatey there, so I run 3.8.1 there too.

jgsogo:
Yes, totally. When building the docker images I was surprised by the latest Ubuntu having an old Python version. In fact I read about it in the Meson repo, latest version cannot be installed in some distros because of Python.

Do you think we should keep py37 and py38 here, just py38 🤷‍♂ ? ideas? suggestions?

Croydon:
imho 3.7+3.8 sounds reasonable

danimtb:
I think we should also include python 3.5. As you said it is very typical to have this version installed by default in the debian 9 and ubuntu 18 distributions (I think these ones are the most popular)

jgsogo:
IMO it makes sense to have more variety of Python versions, so py35 and py38 makes sense as @danimtb  says

PR #156, opened 2020-02-11
Fix CI for Python 2.7

danimtb:
Seems there was a major change in virtualenv when they bumped the version to 20.0.0 - For some reason this caused the pyenv creation to fail only in Python 2.

This is a temporal fix for it until we drop Python 2 support from testing.

---------------

closes #155

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,10 @@ install:
       pyenv activate conan;
     fi
   - pip install --upgrade pip
-  - pip uninstall -y -r <(pip freeze)
+  - |
+    if [[ "$(pip freeze)" ]]; then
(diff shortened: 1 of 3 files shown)


uilianries:
Related issue: https://github.com/conan-io/conan/issues

.travis.yml
```diff
@@ -86,7 +86,7 @@ install:
       pyenv activate conan;
     fi
   - pip install --upgrade pip
-  - pip install tox tox-venv
+  - pip uninstall -y -r <(pip freeze)
```
danimtb: This is done to avoid conflicts with pluggy (requirement of pytest in the travis env and the requirement of tox and tox-env installed by us)

PR #158, opened 2020-02-16
Remove system_requirements exception for libusb 

Croydon:
I would like to create a "virtual package" for libudev in CCI, which relies on system_requirements to install libudev

The rational for whitelisting it in the first place was because @uilianries and I deemed it to be too low-level. Libudev is part of systemd, but I'm not sure how entangled it is with the rest of systemd

Libusb is currently installing libudev via system_requirements.

Moving it into an own recipe allows
  * that other recipes can use it too
  * if we ever decide to build it ourselves we can "just" replace the logic within the recipe

In general, I still firmly believe that system_requirements should be avoided so we should be careful when whitelisting stuff.

// @SSE4 

Blocked by https://github.com/conan-io/conan-center-index/pull/7940


uilianries:
@czoido it's another example related to https://github.com/conan-io/conan-center-index/issues/641

uilianries (dismissed):

SSE4 (dismissed):

jgsogo:
Link: https://github.com/conan-io/conan-center-index/pull/2468

Croydon:
I will update this to remove exception for libusb

All `system` "versions" are allowed by the hooks anyway

Croydon:
This should only get merged after we actually have a system package including libudev, so that the libusb recipe can be updated accordingly

Croydon:
Due to the merge of the default branch, this would now also remove the exception for the `backward-cpp` recipe. Are we okay with this?

uilianries:
> Due to the merge of the default branch, this would now also remove the exception for the `backward-cpp` recipe. Are we okay with this? Okay

uilianries (approved):

PR #161, opened 2020-02-22
Check final newline of selected file types (KB-H041)

### LINKED ISSUE #159: [conan-center] Detect end-of-line for C files on test package
This idea came from https://github.com/conan-io/conan-center-index/pull/910#discussion_r381538273

As C language requires an empty line at the end of the file, and we don't check compiler warnings, we can add a hook which checks for such format.

Referece: https://port70.net/~nsz/c/c11/n1570.html#5.1.1.2p1

/cc @SSE4 @Croydon

madebr:
In response to discussion at https://github.com/conan-io/conan-center-index/pull/910#discussion_r381538273

closes #159

WIKI:

**#KB-H041: "NO FINAL ENDLINE"**

Source files should end with a final endline.
This avoid potential warnings/errors like `no new line at the end of file`.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -43,6 +43,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H034": "TEST PACKAGE - NO IMPORTS()",
              "KB-H037": "NO AUTHOR",
              "KB-H040": "NO TARGET NAME",
+             "KB-H041": "NO FINAL ENDLINE",
             }
 
 
@@ -372,6 +373,30 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                 out.error("CCI uses the name of the package for {0} generator. "
                           "Conanfile should not contain 'self.cpp_info.names['{0}']'. "
                           " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.".format(generator))
+
+    @run_test("KB-H041", output)
+    def test(out):
+        checked_fileexts = ".c", ".cc", ".cpp", ".cxx", ".h", ".hxx", ".hpp", ".py", ".txt", ".yml", ".cmake"
+
+        def _check_final_newline(path):
+            with open(path, "rb") as f:
+                try:
+                    f.seek(-1, 2)  # Move to last byte
+                except OSError:
+                    return  # File is empty ==> ignore
+                if f.read(1) not in (b'\n', '\r'):
+                    out.error("File '{}' does not end with an endline".format(path))
+
+        for root, _, filenames in os.walk(export_folder_path):
+            for filename in filenames:
+                _, fileext = os.path.splitext(filename)
+                if fileext.lower() in checked_fileexts:
+                    _check_final_newline(os.path.join(root, filename))
+
+        config_yml = os.path.join(export_folder_path, os.path.pardir, "config.yml")
+        if os.path.isfile(config_yml):
+            _check_final_newline(config_yml)
+
 
 @raise_if_error_output
 def post_export(output, conanfile, conanfile_path, reference, **kwargs):
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -384,7 +384,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
         def _check_final_newline(path):
             try:
                 last_char = tools.load(path)[-1]
-            except OSError:
+            except (OSError, IndexError):
                 return  # File is empty ==> ignore
             if last_char not in ("\n", "\r"):
                 out.error("File '{}' does not end with an endline".format(path))
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -376,21 +376,23 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H041", output)
     def test(out):
-        checked_fileexts = ".c", ".cc", ".cpp", ".cxx", ".h", ".hxx", ".hpp", ".py", ".txt", ".yml", ".cmake"
+        checked_fileexts = ".c", ".cc", ".cpp", ".cxx", ".h", ".hxx", ".hpp", \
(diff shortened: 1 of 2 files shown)


uilianries:
hmm, some tests are broken, but your test is totally fine.

danimtb:
It is weird to me that it only fails in Python 2, but the error says:

```
E   ERROR: [HOOK - /home/travis/build/conan-io/hooks/tests/test_hooks/conan-center/../../../hooks/conan-center.py] pre_export(): [Errno 22] Invalid argument
```

Is this a pre_export hook or a post_export one?

madebr:
Pre_export.
There is also a test on the `config.yml` file.

hooks/conan-center.py
```diff
@@ -373,6 +374,30 @@ def test(out):
                           "Conanfile should not contain 'self.cpp_info.names['{0}']'. "
                           " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.".format(generator))
 
+    @run_test("KB-H041", output)
+    def test(out):
+        checked_fileexts = ".c", ".cc", ".cpp", ".cxx", ".h", ".hxx", ".hpp", ".py", ".txt", ".yml", ".cmake"
```
danimtb: I think we should not care about cmake, txt, yml or even py files. The origin of the issue was related to C/C++ files and compiler warnings. This will make many of the current recipes fail, and IMO having no newline at the end of the config.yml should be fine

hooks/conan-center.py
```diff
@@ -373,6 +374,30 @@ def test(out):
...
+        def _check_final_newline(path):
+            with open(path, "rb") as f:
+                try:
+                    f.seek(-1, 2)  # Move to last byte
+                except OSError:
+                    return  # File is empty ==> ignore
+                if f.read(1) not in (b'\n', '\r'):
+                    out.error("File '{}' does not end with an endline".format(path))
+
+        for root, _, filenames in os.walk(export_folder_path):
```
danimtb: How do you know the files that are going to be exported? Shouldn't this check be done in a post_export hook?

hooks/conan-center.py
madebr: I added this test because some tools apparently do not behave well when this character is not there. Also, this would avoid github showing the message `no endline` in diffs and when viewing files.

See https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

hooks/conan-center.py
madebr: I copied it to `pre_export` to also include `config.yml`.
For the same reason as I mentioned in my other comment.

hooks/conan-center.py
```diff
@@ -373,6 +374,30 @@ def test(out):
                           "Conanfile should not contain 'self.cpp_info.names['{0}']'. "
                           " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.".format(generator))
 
+    @run_test("KB-H041", output)
+    def test(out):
+        checked_fileexts = ".c", ".cc", ".cpp", ".cxx", ".h", ".hxx", ".hpp", ".py", ".txt", ".yml", ".cmake"
+
+        def _check_final_newline(path):
+            with open(path, "rb") as f:
```
madebr: @danimtb Maybe this is the problem?
I'm opening the file in binary mode (to avoid encoding problems) and test whether the file ends with `b"\r"` or `"b"\n"`.
Maybe python2 does not recognize this

danimtb: it could be, indeed

uilianries: True

hooks/conan-center.py
Croydon: I agree that it would be better to check it for all files, despite that it might be annoying initially

I honestly don't understand why not all editors add per default a newline to files..

madebr: My editor (PyCharm) does not do this either.
But by adding a `.editorconfig` file (see https://github.com/conan-io/conan-center-index/pull/757), it is automatically added.

hooks/conan-center.py
SSE4: you may consider using conan's `tools.load` (which also handles different text file encodings, so it must be more reliable than just opening text file as binary)

hooks/conan-center.py
SSE4: I agree checking all **text** files would be the best. it's not only the C++ throws such warnings.

hooks/conan-center.py
madebr: @SSE4 Good idea!

Btw, did I miss any file type?

SSE4: I don't think so. top list of ext used in repo:
```
 432 py
 418 yml
 316 txt
 170 cpp
 108 patch
  52 c
  18 pack
  18 idx
  11 sample
   9 md
   8 0
   5 h
   3 xml
   3 jpg
   3 cmake
   2 in
   2 exr
   2 am
   2 ac
   2 4
   2 3
   2 2u
   2 2
   1 wav
   1 ttf
   1 tcl
   1 proto
   1 png
   1 pl
   1 pcm
   1 obj
   1 mtl
   1 mak
   1 lock
   1 json
```

madebr:
Great! Now everything is failing!

index

uilianries:
@madebr please update this PR with "closes https://github.com/conan-io/hooks/issues/159"

madebr:
@uilianries Done

uilianries:
Thanks!

madebr:
@uilianries Can I do something about these errors?

uilianries:
They need to be fixed. Your new hook detected other tests as incorrect due the last line. Please, take a look what can be done.

madebr:
@uilianries Thanks, I did not get the tests running before.
Now they do :blush:

SSE4:
@danimtb let's merge?

danimtb:
@madebr please fix conflicts here

madebr:
I've merged with current master.
@jgsogo @danimtb  Please review.

danimtb:
added to the wiki. Thanks!

PR #163, opened 2020-02-24
#160 Do not allow python f-strings

### LINKED ISSUE #160: Do not allow python fstring
This idea came from https://github.com/conan-io/conan-center-index/issues/869

We should not allow Python fstring, as that feature is only available since Python 3.6. For instance:

```python
print(f("Package name: {self.name} and Version: {self.version}")
```

uilianries:
closes #160 

WIKI:

#### **<a name="KB-H042">#KB-H042</a>: "NO F-STRINGS"**

Python f-strings (PEP 498) is only available since Python 3.6.


hooks/conan-center.py
```diff
@@ -373,6 +374,12 @@ def test(out):
                           "Conanfile should not contain 'self.cpp_info.names['{0}']'. "
                           " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.".format(generator))
 
+    @run_test("KB-H042", output)
+    def test(out):
+        for num, line in enumerate(conanfile_content.splitlines(), 1):
+            if 'f"' in line or "f'" in line:
```
SSE4: might be too strict check... e.g. something like that:
```
"""
the code inside the triple quotes with "f"
"""
```
or even:
```
x = "stuff"
```

uilianries: Yes, good catch. I can think in the follow cases:

print(f"{foobar}")
print(
     f"{foobar})
print(
f"{foobar}")

So we have 3 possible prefixes

uilianries:
@SSE4 Now I've applied a regex. I think it fits better, also, I included your examples. Thanks for your ideas.

hooks/conan-center.py
```diff
@@ -43,6 +43,7 @@
              "KB-H034": "TEST PACKAGE - NO IMPORTS()",
              "KB-H037": "NO AUTHOR",
              "KB-H040": "NO TARGET NAME",
+             "KB-H042": "NO F-STRINGS",
```
SSE4: what happened to 41 (also 38, 39?)

uilianries: 41: https://github.com/conan-io/hooks/pull/161

Indeed 38 and 39 were hooks which are merged. I think they were abandoned. I gonna update it to 38. Thanks again.

Croydon: I'm not sure if numbers should be re-used. This could led to confusion. Old hook versions point then to wrong information

jgsogo:
The underlying reason for this hook is to check that the recipe works in some set of python versions, it is not about f-strings itself, right? Then I think that is something to implement with a linter.

I've opened issue https://github.com/conan-io/conan-center-index/issues/1045 because I think it is the way to go in the mid-term, PRs can arrive with non-obvious incompatible Python features (we might need to actually run the recipe with different python versions to check that they behave the same, but it is out-of-scope).

uilianries:
> The underlying reason for this hook is to check that the recipe works in some set of python versions, it is not about f-strings itself, right? Then I think that is something to implement with a linter.
> 
> I've opened issue [conan-io/conan-center-index#1045](https://github.com/conan-io/conan-center-index/issues/1045) because I think it is the way to go in the mid-term, PRs can arrive with non-obvious incompatible Python features (we might need to actually run the recipe with different python versions to check that they behave the same, but it is out-of-scope).

So can we close this one?

jgsogo:
Not yet, implementing that in Conan Center could take some time, there are other things with higher priority. But, once we have the linters, we should remove this one.

hooks/conan-center.py
```diff
@@ -373,6 +374,13 @@ def test(out):
                           "Conanfile should not contain 'self.cpp_info.names['{0}']'. "
                           " Use 'cmake_find_package' and 'cmake_find_package_multi' instead.".format(generator))
 
+    @run_test("KB-H042", output)
+    def test(out):
+        regex = re.compile(r"f[\"'].*[\"']")
```
Artalus: nit: would still trigger on
```
print("""
some "stuff" and some other "stuff"
""")
```

Artalus: Not quite, since now it will skip `x = f"{foo}"` due to `\(` in regex. And then it would probably need to also pass `""" x = f"{foo}" """` and the likings...
The amount of testing for this, however, kinda makes me think that @jgsogo is right in that better approach should be not check for a specific feature (like f-string), but for the recipe being pythonX.Y-compatible as whole.

uilianries: I agree, and I hope those tests soon.

SSE4:
f-strings are allowed, as we support Python 3.6+

PR #167, opened 2020-03-10
[h014] Relax the check for unknown  values

### LINKED ISSUE #165: post build hook with iOS or Android profile build triggers error
I get errors from the post build hooks when creating packages either for iOS or Android.

iOS:
`ERROR: [MATCHING CONFIGURATION (KB-H014)] Packaged artifacts does not match the settings used: os=iOS, compiler=apple-clang (https://github.com/conan-io/conan-center-index/wiki/Error-Knowledge-Base#KB-H014)`

Android:
`ERROR: [MATCHING CONFIGURATION (KB-H014)] Packaged artifacts does not match the settings used: os=Android, compiler=clang (https://github.com/conan-io/conan-center-index/wiki/Error-Knowledge-Base#KB-H014)`

I think this should not be errors, also, I think the combination for os and compiler and be added for iOS and Android, since there are no options other then iOS and Android give you

Minimonium:
There is no need to return an error if the hook doesn't know what to do with the current `os`. 
We can't check all platforms for which people contribute additions to the CCI.

Closes #165


uilianries (dismissed):

memsharded:
I am wondering the purpose of this. This hook is intended for ConanCenter, that it is not running unknown OSs. Could you please clarify? Any feedback @danimtb?

Minimonium:
The idea is that recipes from CCI sometimes run on unsupported by the CCI platforms, but users of such platforms are still willing to contribute changes that would allow supporting such platforms better.

The check right now errors out on unsupported `os` values, which leads to confusion and noise when making such contributions.

danimtb:
Yes, I think it is ok to skip this check for them although those OSs are not supported in CCI

uilianries:
@Minimonium Thank you for your PR. Could you please fix the conflict? So we can merge it.

uilianries (approved):
LGTM

Minimonium:
@uilianries It's been a while. 😃

uilianries (approved):

danimtb (approved):

PR #169, opened 2020-03-12
Support mirrors in conandata

### LINKED ISSUE #168: [conan-center-index] Accept mirrors in conandata.yml
Hi!

The KB-H031 (Conan data reduce), filters package version which contains suffixes (e.g 1.0.0-mirror), resulting in a fail. It would be nice having a specific key only for mirrors:

```yaml
sources:
  "1.2.11":
    sha256: c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
    url: https://zlib.net/zlib-1.2.11.tar.gz
mirrors:
  "1.2.11":
    sha256: c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
    url: https://downloads.sourceforge.net/project/libpng/zlib/1.2.11/zlib-1.2.11.tar.gz
```

Related issue: https://github.com/conan-io/conan-center-index/issues/1095

Regards!

uilianries:
closes #168

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -300,7 +300,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_first_level = ["sources", "patches"]
+        allowed_first_level = ["sources", "patches", "mirrors"]
         allowed_sources = ["url", "sha256", "sha1", "md5"]
         allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5"]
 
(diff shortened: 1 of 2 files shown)


uilianries:
@jgsogo Indeed the solution proposed in https://github.com/conan-io/conan/issues/6349 sounds much better.

danimtb:
@uilianries please consider reimplementation based on the new Conan feature 😄

uilianries:
@danimtb I think we no longer need it. I opened this issue before know about conan-io/conan#6349. As that feature will be available on 1.24, I think this hook is useless now. WDYT?

danimtb:
oh, ok, you are right. In that case, I guess checksums should be the same as the same version of the library, so we will have to repeat them. It is a minor thing, so let's see how it goes and we can improve the conandata interface alter if needed. Thanks!

PR #171, opened 2020-03-13
Test for missing system_libs (KB-H043)

madebr:
Test whether a recipe has all system libraries listed in `cpp_info.system_libs`.
This heuristic only run on Linux and Windows and can only be done for shared libraries.

WIKI:

**#KB-H041: "MISSING SYSTEM LIBS"**

A shared library links to some system libraries which are not set by `cpp_info.system_libs` in `package_info`.
This test can only be run on shared libraries.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1354,7 +1354,6 @@ def _deplibs_from_shlibs(conanfile, out):
                             out.warn("Library dependency '{}' of '{}' has a non-standard name.".format(name, library))
                             continue
                         deplibs.setdefault(match.group(1), []).append(library)
-                x = 0
             else:
                 dep_libs_fn = list(l.replace("NEEDED", "").strip() for l in objdump_output.splitlines() if "NEEDED" in l)
                 for dep_lib_fn in dep_libs_fn:
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1326,7 +1326,6 @@ def _deplibs_from_shlibs(conanfile, out):
                 out.warn("Running objdump on '{}' failed. Is the environment variable OBJDUMP correctly configured?".format(library))
                 continue
             if _get_os(conanfile) == "Windows":
-                open("output.txt", "w").write(objdump_output)
                 for dep_lib_match in re.finditer(r"DLL Name: (.*).dll", objdump_output, re.IGNORECASE):
                     dep_lib_base = dep_lib_match.group(1).lower()
                     deplibs.setdefault(dep_lib_base, []).append(library)
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1079,7 +1079,7 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
 
         missing_system_libs = needed_system_libs.difference(deps_system_libs.union(conanfile_system_libs))
 
-        attribute = 'frameworks' if _get_os(conanfile) else 'system_libs'
+        attribute = 'frameworks' if _get_os(conanfile) == "Macos" else 'system_libs'
         for missing_system_lib in missing_system_libs:
             libs = dict_deplibs_libs[missing_system_lib]
             for lib in libs:
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1,10 +1,13 @@
 import ast
 import collections
+
 import fnmatch
 import inspect
 import os
 import re
 import sys
+import subprocess
+
 from logging import WARNING, ERROR, INFO, DEBUG, NOTSET
 
 import yaml
@@ -52,6 +55,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H039": "NOT ALLOWED ATTRIBUTES",
              "KB-H040": "NO TARGET NAME",
              "KB-H041": "NO FINAL ENDLINE",
+             "KB-H043": "MISSING SYSTEM LIBS",
              "KB-H044": "NO REQUIRES.ADD()",
              "KB-H045": "DELETE OPTIONS",
              "KB-H046": "CMAKE VERBOSE MAKEFILE",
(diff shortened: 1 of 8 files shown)


madebr:
@uilianries 
Using this code, we can also check whether the order of the libraries in `cpp_info.libs` is correct.

madebr:
I think Windows CI fails because it needs the Visual Studio compiler and it is not set in the test section.
(`test_script` section in `appveyor.yml`)

uilianries:
@madebr 

>  Using this code, we can also check whether the order of the libraries in cpp_info.libs is correct.

What do mean by correct? Anyway, it deserves a separated hook.

hooks/conan-center.py
```diff
@@ -573,8 +574,28 @@ def test(out):
             out.error("The conan-center repository doesn't allow Microsoft Visual Studio runtime files.")
             out.error("Found files:\n{}".format("\n".join(bad_files)))
 
-def post_package_info(output, conanfile, reference, **kwargs):
+    @run_test("KB-H043", output)
+    def test(out):
+        if tools.is_apple_os(_get_os(conanfile)):
+            out.info("apple os not supported")
```
uilianries:
```suggestion
```

madebr: You would print nothing at all?

madebr:
@uilianries 

> What do mean by correct?

By correct, I mean correct ordering. Such that dependencies of other libraries come first in the list.

> Do think about header-only libraries? I mean, in terms of test. Some of them contain system_libs.

header-only libraries require another approach.
This pr analyzes shared libraries for its dependencies, which is a fast operation.

For header-only libraries, we'ld need to grep all headers for occurrences of certain includes, which can be time and cpu intensive for larger projects.

e.g. for glibc: (dl, m, pthread and rt)
```python
r"""#\w*include\w*([<"])((dlfcn.h)|(math.h)|(pthread.h)|(time.h))\g<1>"""
```
It can be done, but will be non-perfect.
Compilers might inline some function and don't require linking to an external library.

hooks/conan-center.py
uilianries: most of other hook just skip, no print.

uilianries:
> @uilianries
> 
> > What do mean by correct?
> 
> By correct, I mean correct ordering. Such that dependencies of other libraries come last in the list.
> 
> > Do think about header-only libraries? I mean, in terms of test. Some of them contain system_libs.
> 
> header-only libraries require another approach.
> This pr analyzes shared libraries for its dependencies, which is a fast operation.
> 
> For header-only libraries, we'ld need to grep all headers for occurrences of certain includes, which can be time and cpu intensive for larger projects.
> 
> e.g. for glibc: (dl, m, pthread and rt)
> 
> ```python
> r"""#\w*include\w*([<"])((dlfcn.h)|(math.h)|(pthread.h)|(time.h))\g<1>"""
> ```
> 
> It can be done, but will be non-perfect.
> Compilers might inline some function and don't require linking to an external library.

Okay, let's update it in another occasion.

madebr:
@uilianries 
The tests fail on Appveyor, because it cannot find the Visual Studio installation in the `post_package` hook.
https://ci.appveyor.com/project/ConanOrgCI/hooks/builds/32505216/job/gs8bxp78pu2y82i3#L238
```
E   ERROR: [HOOK - C:\projects\hooks\tests\test_hooks\conan-center\..\..\..\hooks\conan-center.py] post_package(): VS non-existing installation: Visual Studio 14
```
Do you have an idea what might be up?

SpaceIm:
@madebr Does this hook allow shared library to not list system libs if those system libs are not direct dependencies but instead are coming from system libs of dependencies?

madebr:
> @madebr Does this hook allow recipes to not list system libs if those system libs are not direct dependencies but instead are coming from system libs of dependencies?

Yes!

On Linux, it runs `objdump -p $LIBRARY` and parses the output.
I've got this from the `Security` section from the man page of `ldd`.

prince-chrismc:
I've used this tool for windows https://docs.microsoft.com/en-us/cpp/windows/understanding-the-dependencies-of-a-visual-cpp-application?view=vs-2019

There's a link to a GitHub repo with the tool

SSE4:
@madebr I took some look, and it seems like environment is somehow corrupted, e.g. it returns None from:
```
program_files = get_env("ProgramFiles(x86)") or get_env("ProgramFiles")
```
and from:
```
env_var = "vs%s0comntools" % compiler_version
vs_path = os.getenv(env_var)
```
I am pretty sure it's corrupted somewhere inside the test, probably in `_get_environ`? could you double-check that?

SSE4 (dismissed):

uilianries (dismissed):
LGTM

tox.ini
```diff
@@ -23,7 +23,7 @@ setenv =
     coverage: COVERAGE_FILE={toxinidir}/.coverage
     coverage: PYTESTDJANGO_COVERAGE_SRC={toxinidir}/
 
-passenv = PYTEST_ADDOPTS USE_UNSUPPORTED_CONAN_WITH_PYTHON_2 USERNAME
+passenv = PYTEST_ADDOPTS USE_UNSUPPORTED_CONAN_WITH_PYTHON_2 USERNAME ProgramFiles(x86) ProgramFiles
```
madebr, line 26: Nice job @SSE4
How a one line change can fix (almost) everything.

SSE4: it took me two years to find it: https://tox.wiki/en/latest/config.html#conf-passenv 🤦

uilianries: well, today we learned something new at least :nerd_face:

tests/requirements_test.txt
```diff
@@ -6,3 +6,4 @@ pylint==2.10.2
 astroid
 spdx_lookup
 yamllint
+cmake
```
SSE4, line 9:
```suggestion
```

uilianries (approved):
LGTM

hooks/conan-center.py
```diff
@@ -1255,3 +1282,177 @@ def _check_short_paths(conanfile_path, folder_path, max_length_path, output):
...
+                r = re.compile(r)
+                for load_command in load_commands.values():
+                    if load_command.get("cmd") == "LC_LOAD_DYLIB":
+                        name = load_command.get("name", '')
+                        match = re.match(r, name)
+                        if not match:
+                            out.warn("Library dependency '{}' of '{}' has a non-standard name.".format(name, library))
+                            continue
+                        deplibs.setdefault(match.group(1), []).append(library)
+                x = 0
```
SSE4, line 1357:
```suggestion
```

SSE4 (dismissed):

SSE4:
@madebr @uilianries I've added some macOS implementation, it's detecting missing frameworks.
P.S. it should be relatively safe to activate the hook - it only prints warnings. ofcourse, there could be some hard errors (exceptions) from it, but I think we can address them quickly.

hooks/conan-center.py
```diff
@@ -1255,3 +1282,176 @@ def _check_short_paths(conanfile_path, folder_path, max_length_path, output):
...
+                cmd = [objdump, "--section=.idata", "-x", library]
+            else:
+                cmd = [objdump, "-p", library]
+            try:
+                objdump_output = subprocess.check_output(cmd, cwd=conanfile.package_folder).decode()
+            except subprocess.CalledProcessError:
+                out.warn("Running objdump on '{}' failed. Is the environment variable OBJDUMP correctly configured?".format(library))
+                continue
+            if _get_os(conanfile) == "Windows":
+                open("output.txt", "w").write(objdump_output)
```
madebr, line 1329:
```suggestion
```

madebr (commented):
I'm not 100% confident of this pr, but as @SSE4  said.
This hook only emits warnings and bugs can be fixed.

SSE4 (dismissed):

uilianries (approved):

SSE4 (approved):

uilianries (approved):

PR #173, opened 2020-03-18
[conan-center] Versions in conandata should be strings

jgsogo:
Add a check to conan-center hook: versions in conandata should be strings. Comes from https://github.com/conan-io/conan/issues/6697 I thought we were already checking it... have I missed something?


danimtb:
We handle the conversion for versions in the config.yml file and conandata.yml in the pipeline code, but cannot do anything with the internal conversion of Python types in Conan...

jgsogo:
> We handle the conversion for versions in the config.yml file and conandata.yml in the pipeline code, but cannot do anything with the internal conversion of Python types in Conan...

There is nothing we can do in Conan, but in Conan Center it makes sense to force these strings to be strings.

PR #177, opened 2020-04-14
#175 Do not allow requires.add (KB-H044)

### LINKED ISSUE #175: [conan center] check self.requires.add (and self.build_requires.add)
original conversation: https://github.com/conan-io/conan-center-index/pull/1293#discussion_r408189278

some recipes (may) contain `self.requires.add` or `self.build_requires.add`.
these methods aren't currently documented.
they were documented in the past and used across recipes in previous versions of Conan center, but right now they are not part of public API.

/cc @uilianries @jgsogo

uilianries:
closes #175 

Wiki:

#### **<a name="KB-H044">#KB-H044</a>: "NO REQUIRES.ADD()"**

Both `self.requires.add()` and `self.build_requires.add()` have been deprecated in favor of `self.requires()` and `self.build_requires()` which were introduced on Conan 1.x. Both `add()` were introduced during Conan 0.x and since Conan 1.23 they no longer follow the original behavior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,24 +33,24 @@ jobs:
       env: TOXENV=py38-conancurrent
       dist: xenial
 
-    - stage: Linux - Conan 1.21
+    - stage: Linux - Conan 1.23
       name: Python 3.5
       python: 3.5
-      env: TOXENV=py35-conan121
+      env: TOXENV=py35-conan123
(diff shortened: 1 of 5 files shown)


danimtb:
Updated conan-center-index wiki. Thanks!

PR #183, opened 2020-04-24
Check option for deletion

### LINKED ISSUE #179: [conan center] prefer del self.options rather self.options.remove
from https://github.com/conan-io/conan-center-index/pull/1375/files#r409740204
/cc @uilianries 
prefer this syntax:
```
del self.options.darwin_ssl
```
over
```
self.options.remove("darwin_ssl")
```

uilianries:
closes #179

/cc @SSE4 @danimtb 

#### **<a name="KB-H045">#KB-H045</a>: "DELETE OPTIONS"**

The method `self.options.remove` was introduced in Conan 0.x, however Conan 1.x brings a more pythonic way for removing options from your recipe.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -43,6 +43,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H034": "TEST PACKAGE - NO IMPORTS()",
              "KB-H037": "NO AUTHOR",
              "KB-H040": "NO TARGET NAME",
+             "KB-H045": "DELETE OPTIONS",
             }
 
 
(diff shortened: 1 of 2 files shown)


SSE4:
@uilianries I have been thinking about it a lil bit.
with syntax:
```
self.options.remove("shared")
```
it's easy to make typo, e.g.:
```
self.options.remove("shated")
```
while with:
```
del self.options.shared
```
it's much better (at least IntelliSense/editor helps here)

uilianries:
Good catch! I didn't think about the typo case, but it makes total sense as we had real cases in Conan Center Index.

danimtb:
updated wiki

uilianries:
Thanks a lot @danimtb !

Issue #193, opened 2020-05-04
[conan-center] bypass some checks for virtual/system packages


SSE4:
from https://github.com/conan-io/conan-center-index/pull/1535
(same will apply for other system-specific things, such as X11, VDPAU, VAAPI, OpenCL, etc.)
we need to bypass several hooks checks for the virtual (system) packages.
some of them, which fail:

- [IMMUTABLE SOURCES (KB-H010)]
- [PACKAGE LICENSE (KB-H012)]
- [MATCHING CONFIGURATION (KB-H014)]

PR is on the way

PR #194, opened 2020-05-04
do not run certain checks for system packages

SSE4:
closes: #193 

disabled:

- [IMMUTABLE SOURCES (KB-H010)]
- [PACKAGE LICENSE (KB-H012)]
- [MATCHING CONFIGURATION (KB-H014)]

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -418,6 +418,8 @@ def pre_source(output, conanfile, conanfile_path, **kwargs):
 
     @run_test("KB-H010", output)
     def test(out):
+        if conanfile.version == "system":
+            return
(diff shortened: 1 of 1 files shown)


Croydon:
Random idea: Maybe maintaining a whitelist for this is better to avoid that such packages do become the norm?

SSE4:
@Croydon that could be okay, I'd expect just around 20 package of such nature a total. this shouldn't be too annoying to maintain that list within hooks. my initial concern was that we develop a convention (like `library/system`), so version `system` becomes kinda "special" here.

uilianries:
A white list would be nice, but it should be present at the same file, because the idea here is reproducing the same behavior even when running at home

PR #199, opened 2020-05-19
KB-H047 Do not accept non-ascii

uilianries:
This hook prevents any new non-ascii character in any CCI recipe.

In the past, we had a similar problem, but it was related to the author's name in 'author' attribute.

Related to https://github.com/conan-io/conan-center-index/pull/1678

/cc @memsharded 

Wiki:

#### **<a name="KB-H047">#KB-H047</a>: "NO ASCII CHARACTERS"**

According to PEP [263](https://www.python.org/dev/peps/pep-0263/), Unicode literals should only appear in Python code if the encoding is declared on one of the first two lines of the source file. Without such a declaration, any Unicode literal will cause a syntax error for Python 2 interpreters.


tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -1,7 +1,9 @@
+# coding=utf-8
```
uilianries, line 1: Python 2 is able to parse this file when the encoding is not declared

uilianries:
> But, yes, if we don't allow that header, these kind of checks are needed, but we need to check (at least) the conanfile.py inside the test_package too.

Done.

danimtb:
Added to the wiki: https://github.com/conan-io/conan-center-index/wiki/Error-Knowledge-Base#kb-h047-no-ascii-characters

uilianries:
Thanks!

PR #205, opened 2020-06-17
Automatically roll releases

jgsogo:
Same PR as https://github.com/conan-io/examples/pull/65


uilianries:
Can you exclude 123 and 124 from tox file ?

PR #211, opened 2020-07-21
fix KB-H032 SYSTEM REQUIREMENTS

ericLemanissier:
it did not catch the following pattern, just because the presence of "tools.":
installer = tools.SystemPackageTool()
installer.install('libcap-dev')

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -296,7 +296,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H032", output)
     def test(out):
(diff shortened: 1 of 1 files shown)


SSE4:
will it exclude `*/system` packages?

Croydon:
> will it exclude */system packages?

It does not

Back then we still had an allowlist maintained, so only libusb would be allowed

ericLemanissier:
`opengl/system`, `xorg/system` and `glu/system` all use `tools.SystemPackageTool`, this is why they passed CI checks without being rejected. This change would reject them so I added all packages with version `system` into the accept list

PR #213, opened 2020-07-27
Check minimal CMake version for test pkg (KB-H048)

uilianries:
This hooks is the answer for https://github.com/conan-io/conan-center-index/issues/2184

Docs: https://github.com/conan-io/conan-center-index/pull/2284

- CMake +3 is mandatory for test_package/CMakeLists.txt
- If CXX_STANDARD or CMAKE_CXX_STANDARD is declared, CMake +3 is mandatory

closes https://github.com/conan-io/conan-center-index/issues/2184

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -48,6 +48,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H045": "DELETE OPTIONS",
              "KB-H046": "CMAKE VERBOSE MAKEFILE",
              "KB-H047": "NO ASCII CHARACTERS",
+             "KB-H048": "CMAKE VERSION REQUIRED",
             }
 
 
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -461,6 +462,24 @@ def check_for_verbose_flag(cmake_path):
             if os.path.exists(cmake_path):
                 check_for_verbose_flag(cmake_path)
 
+    @run_test("KB-H048", output)
+    def test(out):
+        dir_path = os.path.dirname(conanfile_path)
+        cmake_test_pkg = os.path.join(dir_path, "test_package", "CMakeLists.txt")
+        if os.path.isfile(cmake_test_pkg):
+            cmake_content = tools.load(cmake_test_pkg)
+            if "cmake_minimum_required(version 2" in cmake_content.lower():
```
SSE4, line 471: what about `3` and `3.0`, or is it unlikely to be encountered?

SSE4: also, `1`

uilianries: that's a good question actually, here we only prevent CMake 2, but I didn't think about 3.1. Do you think we should force 3.1 only? I can't remember if we have any edge case which requires >3.1, maybe using macros?

danimtb:
What about `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`? That one requires 3.4 and is usually forgotten

danimtb:
> What about CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS? That one requires 3.4 and is usually forgotten

Ok, now I see this only for test_package so it doesn't apply. Please dismiss my comment.

uilianries:
> > What about CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS? That one requires 3.4 and is usually forgotten
> 
> Ok, now I see this only for test_package so it doesn't apply. Please dismiss my comment.

True, but we can consider it for another hook, why not.

PR #218, opened 2020-08-18
Remove extra patches (keep revisions untouched)

jgsogo:
When a PR is adding a patch for some revision that doesn't modify the `conanfile.py`, now we are creating a new recipe revision because the exported patch is added to the manifest. This PR removes the patches that are exported but not listed in the `conandata.yml` file for the exported version.

This PR uses lot of assumptions: patches are exported to a `patches` directory, they are listed in the `conandata.yml` file,... so this hook can fail due to different reasons, please, pay attention in the review and check that every failure is meaningful.

wait for https://github.com/conan-io/hooks/pull/221


danimtb:
We have to resume work here regarding patching

jgsogo:
Now, I'm seeing this pattern in many recipes:

```python
    def export_sources(self):
        self.copy("CMakeLists.txt")
        for patch in self.conan_data.get("patches", {}).get(self.version, []):
            self.copy(patch["patch_file"])
```

The recipe is taking care of exporting only the patches it is going to use. Maybe this hook is no longer needed.

danimtb:
> Now, I'm seeing this pattern in many recipes:
> 
> ```python
>     def export_sources(self):
>         self.copy("CMakeLists.txt")
>         for patch in self.conan_data.get("patches", {}).get(self.version, []):
>             self.copy(patch["patch_file"])
> ```
> 
> The recipe is taking care of exporting only the patches it is going to use. Maybe this hook is no longer needed.

That's great then! Probably we have to document this as good practice, so it is easy to check by reviewers or maybe have a hook check that makes sure that the files listed in the reduced conandata are exactly the same ones exported.

prince-chrismc:
> Probably we have to document this as good practice

https://github.com/conan-io/conan-center-index/pull/11982 👍

jgsogo:
I'm going to close this PR, I think it is better to be explicit (those extra lines in the recipe) than implicit/magic (hook applying changes).

In case we can implement a hook for this, it is better to CHECK that those files are removed instead of MODIFYING whatever is exported.

PR #219, opened 2020-08-18
[conan-center] KB-H049: CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS requires CMake 3.4

uilianries:
We have PRs in CCI which requires CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS, but we know, it requires CMake 3.4 at least: https://cmake.org/cmake/help/v3.4/variable/CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS.html

Docs: https://github.com/conan-io/conan-center-index/pull/2593

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -495,6 +495,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                         out.error("The CMake definition {} requires CMake 3.4 at least. Update "
                                 "CMakeLists.txt to 'cmake_minimum_required(VERSION 3.4)'."
                                 .format(cmake_def))
+                        break
 
 
 @raise_if_error_output
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -49,6 +49,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H046": "CMAKE VERBOSE MAKEFILE",
              "KB-H047": "NO ASCII CHARACTERS",
              "KB-H048": "CMAKE VERSION REQUIRED",
+             "KB-H049": "CMAKE EXPORT ALL SYMBOLS",
             }
 
 
@@ -479,6 +480,17 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                "cxx_standard" in cmake_content.lower():
                 out.error("The CMake definition CXX_STANDARD requires CMake 3.1 at least."
                           " Update to 'cmake_minimum_required(VERSION 3.1)'.")
+
+    @run_test("KB-H049", output)
+    def test(out):
+        dir_path = os.path.dirname(conanfile_path)
+        cmake_path = os.path.join(dir_path, "CMakeLists.txt")
+        if os.path.isfile(cmake_path):
+            cmake_content = tools.load(cmake_path)
+            match = re.search(r"cmake_minimum_required\s?\(VERSION (\d?\.?\d?\.?\d+)\)", cmake_content, re.I)
+            if match and tools.Version(match.group(1)) < "3.4" and "WINDOWS_EXPORT_ALL_SYMBOLS" in cmake_content:
+                out.error("The CMake definition WINDOWS_EXPORT_ALL_SYMBOLS requires CMake 3.4 at least."
+                          " Update to 'cmake_minimum_required(VERSION 3.4)'.")
 
 
 @raise_if_error_output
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -480,6 +481,19 @@ def test(out):
...
+    @run_test("KB-H049", output)
+    def test(out):
+        dir_path = os.path.dirname(conanfile_path)
+        cmake_path = os.path.join(dir_path, "CMakeLists.txt")
+        if os.path.isfile(cmake_path):
+            cmake_content = tools.load(cmake_path)
+            match = re.search(r"cmake_minimum_required\s?\(VERSION (\d?\.?\d?\.?\d+)\)",
+                              cmake_content, re.I)
+            if match and tools.Version(match.group(1)) < "3.4" and \
+               "CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS" in cmake_content:
```
SSE4, line 493: I'd check just `WINDOWS_EXPORT_ALL_SYMBOLS`, as it can be set on target, not globally

hooks/conan-center.py
```diff
@@ -480,6 +481,19 @@ def test(out):
...
+                out.error("The CMake definition CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS requires CMake 3.4"
+                          " at least. Update to 'cmake_minimum_required(VERSION 3.4)'.")
```
jgsogo, line 495: Tell the file where you found the `WINDOWS_EXPORT_ALL_SYMBOLS`, at least there are two `CMakeLists.txt` files (yes, we expect/check this error only from the one building the package, but it is useful information anyway).

hooks/conan-center.py
uilianries: Thanks! I was not sure if WINDOWS_EXPORT_ALL_SYMBOLS was really important for this hook. I'll update.

hooks/conan-center.py
uilianries: Good point! I'll update it.

uilianries:
@jgsogo @SSE4 I just updated the hook. Thanks for your comments!

hooks/conan-center.py
```diff
@@ -480,6 +481,20 @@ def test(out):
...
+        cmake_path = os.path.join(dir_path, "CMakeLists.txt")
+        if os.path.isfile(cmake_path):
+            cmake_content = tools.load(cmake_path)
+            match = re.search(r"cmake_minimum_required\s?\(VERSION (\d?\.?\d?\.?\d+)\)",
+                              cmake_content, re.I)
+            if match and tools.Version(match.group(1)) < "3.4":
+                for cmake_def in ["WINDOWS_EXPORT_ALL_SYMBOLS", "CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"]:
+                    out.error("The CMake definition {} requires CMake 3.4 at least. Update "
+                              "CMakeLists.txt to 'cmake_minimum_required(VERSION 3.4)'."
+                              .format(cmake_def))
```
jgsogo, line 496: 🤔  I think we need to check if the string `WINDOWS_EXPORT_ALL_SYMBOLS` is found inside `cmake_content`... maybe it is just me that cannot see it, but... 😄

uilianries: It works too, but the logic looks wrong of course, updating in 1 .. 2.. 3..

uilianries: Done!

hooks/conan-center.py
```diff
@@ -480,6 +481,21 @@ def test(out):
...
+    @run_test("KB-H049", output)
+    def test(out):
+        dir_path = os.path.dirname(conanfile_path)
+        cmake_path = os.path.join(dir_path, "CMakeLists.txt")
+        if os.path.isfile(cmake_path):
+            cmake_content = tools.load(cmake_path)
+            match = re.search(r"cmake_minimum_required\s?\(VERSION (\d?\.?\d?\.?\d+)\)",
+                              cmake_content, re.I)
+            if match and tools.Version(match.group(1)) < "3.4":
+                for cmake_def in ["WINDOWS_EXPORT_ALL_SYMBOLS", "CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"]:
```
jgsogo, line 493:
```suggestion
                for cmake_def in ["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS", "WINDOWS_EXPORT_ALL_SYMBOLS"]:
```

hooks/conan-center.py
```diff
@@ -480,6 +481,21 @@ def test(out):
...
+                    if cmake_def in cmake_content:
+                        out.error("The CMake definition {} requires CMake 3.4 at least. Update "
+                                "CMakeLists.txt to 'cmake_minimum_required(VERSION 3.4)'."
+                                .format(cmake_def))
```
jgsogo, line 497:
```suggestion
                                .format(cmake_def))
                        break
```

uilianries:
@jgsogo Done

jgsogo (approved):

Issue #217, opened 2020-08-17
[conan-center] Create check that static libraries are the default


Croydon:
Why we need this hook, e.g.: https://github.com/conan-io/conan-center-index/issues/2566

We already de-facto enforce through reviews that static libraries should be the default option

I propose that we are adding a check to the hook enforcing this automatically with an allowlist for special cases where this might be unwanted

Recipes who don't support shared building won't have the `shared` option and the hook would not complain

Croydon:
Recipes who currently set 

> default_options = {"shared": True}

backward-cpp
botan
glib
libb2
libepoxy
libsigcpp
paho-mqtt-c
tbb
xkbcommon

Croydon:
@ericLemanissier We need to add glib to an allowlist too because it does not support static on Windows, right?

ericLemanissier:
Yes https://gitlab.gnome.org/GNOME/glib/-/issues/692

uilianries:
Backward-cpp does not require shared by default: https://github.com/conan-io/conan-center-index/pull/2598

Croydon:
@uilianries Thanks for helping!

jgsogo:
Thank you two for taking care of this. I also think this is a good default 🙌

uilianries:
Backward-cpp: conan-io/conan-center-index#2598
Blake2 (libb2): https://github.com/conan-io/conan-center-index/pull/2601
libsigcpp: https://github.com/conan-io/conan-center-index/pull/2603
libepoxy: https://github.com/conan-io/conan-center-index/pull/2602
Botan: https://github.com/conan-io/conan-center-index/pull/2614
xkbcommon: https://github.com/conan-io/conan-center-index/pull/2613


TBB: does not officially support static lib: https://github.com/oneapi-src/oneTBB/issues/221#issuecomment-633561961

Croydon:
All done

Thanks everyone 👏

PR #220, opened 2020-08-18
check default shared option value

uilianries:
Add a new check to validate options.shared when is True by default. It's a rare case, and most of time is a mistake.

closes https://github.com/conan-io/hooks/issues/217
Docs: https://github.com/conan-io/conan-center-index/pull/2615

To understand the allowlist, read the issue, there are more details there.

/cc @Croydon

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -509,16 +509,13 @@ def post_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H050", output)
     def test(out):
-        # TODO: Add current allowlist from CCI
-        if conanfile.name in ["glib", "paho-mqtt-c"]:
+        if conanfile.name in ["glib", "paho-mqtt-c", "tbb"]:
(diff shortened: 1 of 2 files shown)


jgsogo (approved):
LGTM 👍

danimtb (approved):

hooks/conan-center.py
```diff
@@ -296,7 +297,7 @@ def test(out):
 
     @run_test("KB-H032", output)
     def test(out):
-        if conanfile.name in ["libusb"] or conanfile.version == "system":
+        if conanfile.name in ["libusb", "backward-cpp"] or conanfile.version == "system":
```
ericLemanissier, line 300: @uilianries do you recall why `backward-cpp` was added to the `system_requirements` whitelist ? I don't understand why the packages installed in https://github.com/conan-io/conan-center-index/blob/41d4fbb8c218f9d938ccb8b9d888bff52002e62e/recipes/backward-cpp/all/conanfile.py#L63-L89 don't have a dedicated package

PR #223, opened 2020-08-23
check if version is missing from config.yml

SSE4:
this is a common mistake to add a new version to the `conandata.yml`, but forget to add it to the `config.yml`.
as result, CI doesn't build the newly added version, and we don't know if it actually passed the CI.
it's very easy to miss during the review, so I've decided to add an automatic check for human error.

NOTE: it's not needed to add the opposite case - version is added to the `config.yml` but not to the `conandata.yml`, as it would fail anyway during the attempt to fetch sources.

docs: https://github.com/conan-io/conan-center-index/pull/2640

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -49,7 +49,8 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H048": "CMAKE VERSION REQUIRED",
              "KB-H049": "CMAKE WINDOWS EXPORT ALL SYMBOLS",
              "KB-H050": "DEFAULT SHARED OPTION VALUE",
-             "KB-H051": "DEFAULT OPTIONS AS DICTIONARY"
+             "KB-H051": "DEFAULT OPTIONS AS DICTIONARY",
+             "KB-H052": "CONFIG.YML HAS NEW VERSION"
              }
 
 
@@ -133,6 +134,12 @@ def run_test(kb_id, output):
             raise e
 
     return tmp
+
+
+def load_yml(path):
+    if os.path.isfile(path):
+        return yaml.safe_load(tools.load(path))
+    return None
 
 
 @raise_if_error_output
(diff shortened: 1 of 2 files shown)


tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -880,3 +880,37 @@ class AConan(ConanFile):
...
+        versions:
+          1.0:
+            folder:all
+          2.0:
+            folder:all
+        """)
+        tools.save("config.yml", content=config)
+        output = self.conan(['export', 'all', 'name/version@user/test'])
+        self.assertNotIn("ERROR: [CONFIG.YML HAS NEW VERSION (KB-H052)] The version \"2.0\" exists in",
+                         output)
```
jgsogo, line 916: Why the ERROR now? Is it because it is reading the version as a float?

SSE4:
It reads version as is (as strong). Now it checks there is no error after
version was added to the config.

jgsogo (approved):
`assertNotIn` 🤦   ...too early in the morning...

danimtb:
I am not sure about having this check in the hook, as the config.yml file is something specific to conan-center-index. Maybe this is a check we should do in the pipeline itself. wdyt @jgsogo?

SSE4:
@danimtb but this is conan-center hook, right? anyway, if there is no config.yml, the check will not run, so it's harmless

jgsogo:
I agree that hooks should use `config.yml` because it is a file that is strange to Conan, in that sense, it would be better to add this check (and any other involving `config.yml` to the CI), but it is really easy to implement it here, these are **conan-center** hooks and it is useful (in the future, if we add this check to the CI we can remove it from here).

hooks/conan-center.py
```diff
@@ -509,6 +514,32 @@ def test(out):
...
+        if 'versions' not in config_yml:
+            return
+
+        if 'sources' not in conandata_yml:
+            return
+
+        versions_conandata = conandata_yml['sources'].keys()
+        versions_config = config_yml['versions'].keys()
+
+        for version in versions_conandata:
```
uilianries, line 537: The opposite also can occur. I can remember one or two cases where the user created a config.yml with more versions but the conandata.yml had only one, like preparing for the future.

SSE4: but didn't it fail anyway? because there are no sources in conandata, so source method should have failed

Croydon:
conan-center hooks are already strongly CCI specific anyway, e.g. we check if the homepage is set to CCI

And for the sake of transparency and open source I would actually suggest to implement even more into the hooks =)

danimtb (approved):

hooks/conan-center.py
jgsogo: We can add the opposite check in the future (even for this same hook).

Issue #176, opened 2020-04-14
[conan center] generic check for undocummented APIs


SSE4:
is it somehow possible more or less generic checks for undocumented APIs in recipes?
we had some issues with it previously, e.g.:
https://github.com/conan-io/conan-center-index/pull/1293#discussion_r408189278
https://github.com/conan-io/conan-center-index/pull/1284
maybe we can avoid such kind of issues
possible methods:
- check for black/white -listed namespaces/imports? (e.g. `conans.client.build` might be black-listed)
- somehow export list of public APIs from Conan codebase or Conan docs?

suggestions are welcomed

/cc @uilianries @jgsogo

jgsogo:
Probably this is something to do using a linter, together with other checks that are currently implemented in the hooks. Probably it is the easiest way and it should cover 100% of the scenarios, other approaches with regex/find always have exceptions.

jgsogo:
Not sure if this should be enough with a file in `conans/pylint_plugin.py` or there would be some specific errors here and we will need to maintain another file in this repo

jgsogo:
btw, I totally support using regex/find strategies while we add the linter to the pipeline (this is something I've already talked with @uilianries in some PRs here but we couldn't find time to do it) 😓

SSE4:
@jgsogo  do you have an example of how we can implement such a check with `conans/pylint_plugin.py`?
will we need to maintain black/white list of functions in Conan codebase?
will we need to somehow mark API functions in Conan codebase?
can we just assume regular functions are public API, but something with one or two underscores is private?

jgsogo:
Maybe we need a plain `pylint_plugin` adding only the objects and methods we allow in ConanCenter and run pylint without in a Python environment without Conan installed

PR #226, opened 2020-08-26
- check private imports

SSE4:
related https://github.com/conan-io/hooks/issues/176

check for private APIs, disallowed examples:
```
from conans.model.version import Version
from conans.client.tools import msvs_toolset
```
allowed examples:
```
from conans.tools import Version
from conans.errors import ConanInvalidConfiguration
```


danimtb (approved):

SSE4:
docs: https://github.com/conan-io/conan-center-index/pull/2662

uilianries:
That's a great hook! At least two times I needed to restrict it in CCI during a review!

hooks/conan-center.py
```diff
@@ -552,6 +553,32 @@ def test(out):
                           'version "{}".'.format(version, conandata_path, config_path, config_path,
                                                  version))
 
+    @run_test("KB-H053", output)
+    def test(out):
+        def _is_private_import(line):
+            allowed_list = ["tools", "errors"]
+            for pattern in ["from conans.", "import conans."]:
```
uilianries, line 560: well, I can import conans and call conans.user.something :smiling_imp:

SSE4: I don't think we should go too deep, just prevent simple cases.
I still can do alias imports and other hacks, but that's beyond the goal.

uilianries: I agree!

jgsogo (approved):
Like it. It will hit some recipes and they will become better recipes 😄

jgsogo:
@SSE4 we need to fix a conflict here

jgsogo:
Now a different conflict, after reverting KB053, nevertheless I'll add this one as KB054 (so we don't modify https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#kb-h053-invalid-symlinks), and we will see if we can merge back KB053 again

SSE4:
what? I've just solved conflict 15 minutes ago... how come it keeps conflicting every time?

hooks/conan-center.py
```diff
@@ -50,7 +50,9 @@
              "KB-H049": "CMAKE WINDOWS EXPORT ALL SYMBOLS",
              "KB-H050": "DEFAULT SHARED OPTION VALUE",
              "KB-H051": "DEFAULT OPTIONS AS DICTIONARY",
-             "KB-H052": "CONFIG.YML HAS NEW VERSION"
+             "KB-H052": "CONFIG.YML HAS NEW VERSION",
+             "KB-H053": "PRIVATE IMPORTS",
+             "KB-H054": "INVALID SYMLINKS"
```
jgsogo, line 55:
```suggestion
             "KB-H053": "INVALID SYMLINKS"
             "KB-H054": "PRIVATE IMPORTS",
```

Right now it should go in this ordering, unless we wait after the _symlinks_ issue is resolved

hooks/conan-center.py
```diff
@@ -50,7 +50,9 @@
              "KB-H049": "CMAKE WINDOWS EXPORT ALL SYMBOLS",
              "KB-H050": "DEFAULT SHARED OPTION VALUE",
              "KB-H051": "DEFAULT OPTIONS AS DICTIONARY",
-             "KB-H052": "CONFIG.YML HAS NEW VERSION"
+             "KB-H052": "CONFIG.YML HAS NEW VERSION",
+             "KB-H053": "INVALID SYMLINKS"
```
jgsogo, line 54:
```suggestion
             "KB-H053": "INVALID SYMLINKS",
```

jgsogo (approved):

PR #228, opened 2020-09-02
KB-H054: Check for targets when building header-only

### LINKED ISSUE #227: [conan-center] Check CMake target for test_package when building hearder-only
This hook is related to https://github.com/conan-io/conan-center-index/pull/2669#discussion_r478446708

Basically, when a recipe is header-only, and the test_package imports the module using cmake (find_package), we have to use `conan_basic_setup(TARGET)` instead of `conan_basic_setup()`

uilianries:
fixes #227 
Docs: https://github.com/conan-io/conan-center-index/pull/2794

If a header-only recipe uses targets in test_package, it should use conan_basic_setup(TARGETS) too.

For ${CONAN_LIBS} the global conan_basic_setup() should be okay.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -50,7 +50,8 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H049": "CMAKE WINDOWS EXPORT ALL SYMBOLS",
              "KB-H050": "DEFAULT SHARED OPTION VALUE",
              "KB-H051": "DEFAULT OPTIONS AS DICTIONARY",
-             "KB-H052": "CONFIG.YML HAS NEW VERSION"
+             "KB-H052": "CONFIG.YML HAS NEW VERSION",
+             "KB-H054": "HEADER ONLY AND CMAKE TARGET"
              }
 
 
@@ -553,6 +554,17 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                           ' Please update "{}" to include newly added '
                           'version "{}".'.format(version, conandata_path, config_path, config_path,
                                                  version))
+
+    @run_test("KB-H054", output)
+    def test(out):
+        dir_path = os.path.dirname(conanfile_path)
+        cmake_path = os.path.join(dir_path, "test_package", "CMakeLists.txt")
+        if header_only and os.path.isfile(cmake_path):
+            content = tools.load(cmake_path).lower()
+            if "conan_basic_setup()" in content and "${conan_libs}" not in content:
+                out.error("The test_package/CMakeLists.txt uses 'conan_basic_setup()', but the "
+                          "recipe is header only. Add 'TARGETS' to 'conan_basic_setup()'.")
+
 
 @raise_if_error_output
 def post_export(output, conanfile, conanfile_path, reference, **kwargs):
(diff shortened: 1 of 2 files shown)


SSE4:
I am not sure I understand the check...
if I am using `conan_basic_setup`, but not using `${CONAN_LIBS}`, it doesn't mean I am using targets, right?
e.g. I may use:
- `${CONAN_LIBS_MYLIB}`
- `${CONAN_LIBS_DEBUG}`/`{CONAN_LIBS_RELEASE}`
- `include_directories(${CONAN_INCLUDE_DIRS_MYLIB}`
- `cmake_find_package` or `cmake_find_package_multi`
- other ways...

in general, I don't understand why would I need `${CONAN_LIBS}` for header-only at all - if there are no libraries, I don't need to link with libraries...

uilianries:
>  if I am using conan_basic_setup, but not using ${CONAN_LIBS}, it doesn't mean I am using targets, right?

Correct, I didn't think about your cases because they are rare.

> in general, I don't understand why would I need ${CONAN_LIBS} for header-only at all - if there are no libraries, I don't need to link with libraries...

I agree. So should we consider Targets only header only? We could filter "\${CONAN.*}"

SSE4:
sorry, I still don't get the check...
e.g. we have a recipe with uses `conan_basic_setup()` and `CONAN_PKG::mylib`, right?
but it will fail anyway, as `CONAN_PKG::mylib` is not defined without targets.
so it's not something to be caught by the hook, as CI will complain in any case...

jgsogo:
Yes, I think there is too much variability here to make a hook. It would be easier if we think about the future: the `cmake[_multi]` generators will likely be deprecated, and we should promote `cmake_find_package[_multi]` together with the new toolchains approach (if we don't see any blocker, it provides a cleaner integration with the build-system).

Given that, as a hook, we could check for `conan_basic_setup(TARGETS)` and no `CONAN_PKG::`. This would be easy to check, but probably it is too early to add this kind of hook because many recipes still use the `cmake` + `CONAN_LIBS` or the `CONAN_PKG::` targets and we don't want to break all of them.

wdyt?

SSE4:
to be honest, I don't see value in both checks, as:
1. it's legal to use `conan_basic_setup(TARGETS)` without `CONAN_PKG::mylib`
2. it's legal to use `conan_basic_setup()` without `${CONAN_LIBS}`

either way, the check doesn't provide much value and appears to be very fragile, as it may reject a correct code.
and I think we're going away from `conan_basic_setup` in any form in the near future anyway.

uilianries:
> sorry, I still don't get the check...

@SSE4 Avoid header only + ${CONAN_LIBS}, or any other variable.

And yes, it difficult to filter all possibilities, but I don't people using exotic variables, only CONAN_LIBS in fact.

SSE4:
still all cases are valid:
1. header only + `conan_basic_setup()` + `CONAN_LIBS`
2. header only + `conan_basic_setup()` without `CONAN_LIBS`
3. header only + `conan_basic_setup(TARGETS)` + `CONAN_LIBS`
4. header only + `conan_basic_setup(TARGETS)` without `CONAN_LIBS`

(same actually for non-header only)

if I link to `CONAN_LIBS` but I am header only, the variable `CONAN_LIBS` will be empty, so it's harmless to use it.
if I don't link to `CONAN_LIBS` - still harmless, as header only - means no libraries.

am I missing something? all these examples are legal, so there is nothing to reject

uilianries:
I agree, closing this one

PR #250, opened 2020-11-11
Ignore Cyrus SASL

uilianries:
Related to https://github.com/conan-io/conan-center-index/pull/3466

/cc @SpaceIm

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -606,7 +606,7 @@ def post_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H050", output)
     def test(out):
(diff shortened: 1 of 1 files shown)


Croydon (approved):

SpaceIm:
We could keep `shared=False` as default value, but it would require `with_gssapi=False` as default value. Need to check what's the default value upstream, and if `shared=False` and `with_gssapi=True` is really forbidden.

SpaceIm:
Not required anymore, `cyrus-sasl:with_gssapi` has been disabled by default, since underlying recipe is not available in CCI. Therefore cyrus-sasl shared option is now False by default.

PR #245, opened 2020-10-19
KB-H053: allow custom generators

ericLemanissier:
conans.model.Generator is explicitly documented in https://docs.conan.io/en/latest/howtos/custom_generators.html

see https://github.com/conan-io/conan-center-index/pull/1759#issuecomment-711129395

PR #253, opened 2020-11-24
Fix hook after PR #245

jgsogo:
https://github.com/conan-io/hooks/pull/245 was not working as expected 🙈


jgsogo:
Let's merge this PR after the CI passes

ping @ericLemanissier

ericLemanissier:
As you wish, I don't mind using `from conans.model import Generator` or `from conans import model.Generator` (if it's valid python ?)

ericLemanissier (approved):

Issue #254, opened 2020-11-24
Check that requires and requirements() do not exist simultaneously


mathbunnyru:
As @SpaceIm suggested in the PR review:
https://github.com/conan-io/conan-center-index/pull/3681#pullrequestreview-537524674

> requires() should be called in requirements() + it's better to put requirements close to each others (harder to understand dependencies in the recipe if they are split in requires attribute and requirements())

SpaceIm:
Maybe a suggestion for https://github.com/conan-io/conan-center-index/pull/3614?

mathbunnyru:
@SpaceIm I added a suggestion.

uilianries:
It sounds good!

PR #261, opened 2020-12-23
Do not accept two ways for requirements

uilianries:
closes https://github.com/conan-io/hooks/issues/254

/cc @SpaceIm @mathbunnyru

Wiki:

#### **<a name="KB-H055">#KB-H055</a>: "SINGLE REQUIRES"**

Both `requirements()` and `self.requires` should not be used together at same recipe. The duality creates a heterogeneous way of solving dependencies, making it difficult to review and susceptible to prone errors.


hooks/conan-center.py
```diff
@@ -583,6 +584,13 @@ def _check_private_imports(filename, content):
             test_package_content = tools.load(test_package_path)
             _check_private_imports("test_package/conanfile.py", test_package_content)
 
+    @run_test("KB-H054", output)
+    def test(out):
+        if hasattr(conanfile, "requires") and callable(getattr(conanfile, "requirements", None)):
+            out.error("Both 'requires' attribute and 'requirements' method should not be "
```
SSE4, line 590: hm, to clarify, why not? what happens if you declare both, which one takes priority?

uilianries: I guess self.requires comes first: https://github.com/conan-io/conan/blob/develop/conans/model/conan_file.py#L47

However, I see no motivation to use both, otherwise we would need to check places instead of one.

mathbunnyru, line 590:
```suggestion
            out.error("Both 'requires' attribute and 'requirements()' method should not be "
```

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -951,3 +952,27 @@ def test_private_import(self):
...
+        class MockRecipe(ConanFile):
+            requires = "foo/0.1.0"
+
+            def requirements(self):
+                self.requires("bar/0.1.0")
+        """)
+
+        tools.save('conanfile.py', content=conanfile)
+        output = self.conan(['export', '.', 'name/version@user/test'])
+        self.assertIn("[SINGLE REQUIRES (KB-H054)] Both 'requires' attribute and 'requirements' "
```
mathbunnyru, line 968:
```suggestion
        self.assertIn("[SINGLE REQUIRES (KB-H054)] Both 'requires' attribute and 'requirements()' "
```

mathbunnyru (changes requested):
LGTM, but I think having parentheses makes it a bit more clear.

mathbunnyru:
Do we want to check build_requires as well?

mathbunnyru (approved):
Good job 👍

SSE4 (approved):

jgsogo:
Please, do not merge it yet... if some recipes are going to fail (https://github.com/conan-io/conan-center-index/issues/2232). First, we would like to fix internal jobs (jobs building missing configurations behind the scenes) so they don't fail due to hook-validation, then we can introduce as many hooks as we want that fail in PRs.

Let's wait after 🎅 

Thanks!

mathbunnyru:
@jgsogo is it already possible to merge this PR?

danimtb (approved):

jgsogo (approved):
Please, @uilianries , fix conflicts

uilianries:
@jgsogo Done, I needed to update hook's id too.

mathbunnyru (approved):

jgsogo (approved):
I hope it doesn't break more than needed 😅

PR #265, opened 2021-01-11
Fix tests (Conan 1.33 modifies ConanFile base class)

jgsogo:
PR https://github.com/conan-io/conan/pull/8243 broke this test, we need to use a different field now.

### DIFF
--- a/tests/test_hooks/test_recipe_linter.py
+++ b/tests/test_hooks/test_recipe_linter.py
@@ -92,7 +92,7 @@ class RecipeLinterTests(ConanClientTestCase):
 
             class Recipe(ConanFile):
                 def build(self):
-                    self.output.info(self.source_folder)
+                    self.output.info(self.conan_data)
         """)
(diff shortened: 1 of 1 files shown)


danimtb:
it should only fail with Conan V2 mode active, right?

jgsogo:
I don't see how Conan v2 plays a role here. Probably I'm missing some information here.

ericLemanissier:
I hope conanfiles will still have a source_folder attribute in conan 1.33 with V2 mode disabled. Otherwise the world will break very fast!

memsharded:
> I hope conanfiles will still have a source_folder attribute in conan 1.33 with V2 mode disabled. Otherwise the world will break very fast!

Yes, it does, but it is a property, not a dynamic attribute anymore. This is why it is being removed from the linter.

danimtb (approved):

SSE4 (approved):

uilianries (approved):

PR #266, opened 2021-01-19
 [KB-H050] add vulkan-loader to whitelist

SpaceIm:
`vulkan-loader` should always be shared (big warning in CMakeLists: https://github.com/KhronosGroup/Vulkan-Loader/blob/aed40d52837dba3850a27c61529de6fa25979d19/CMakeLists.txt#L45), except on Apple os where it's allowed to build it as a static lib.

Will allow https://github.com/conan-io/conan-center-index/pull/4308 to pass CI in CCI


danimtb (approved):

uilianries (approved):

ericLemanissier:
this may not be needed any more, https://github.com/conan-io/conan-center-index/pull/4308 is merged

SpaceIm:
Due to a trick, but we might want to set default option to shared unconditionally.

PR #273, opened 2021-02-19
check the existence of declared libraries

ericLemanissier:
This is a proposition, preventing problems like https://github.com/conan-io/conan-center-index/pull/4610 from happening.
This hook checks 3 things:

1. if some libdir contain no library files at all (warning)
2. if some libdir contain libraries, but none of these are declared in the libs attribute (warning)
3. if some libraries declared in libs attribute cannot be found in the associated libdirs (error)

If you are ok with this idea, I'll add unit tests and documentation

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -52,6 +52,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H051": "DEFAULT OPTIONS AS DICTIONARY",
              "KB-H052": "CONFIG.YML HAS NEW VERSION",
              "KB-H053": "PRIVATE IMPORTS",
+             "KB-H054": "LIBRARY DOES NOT EXIST"
              }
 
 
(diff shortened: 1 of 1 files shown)


uilianries:
Sounds good to me. If you explicitly declared a library, but it is not packaged, it's a prone error.

ericLemanissier:
Thanks. Is there some kind of existing facility to test a hook on all CCI recipes ?

uilianries:
Unfortunately not, but I would say to not use CCI. Any change there can break a test here. You could a recipe from there for testing of course, but downloading real recipes from there can be unstable. Or, maybe, using revisions it could work.

ericLemanissier:
I just made a dumb loop trying `conan install` with hooks on all packages on CCI, currently running at https://github.com/ericLemanissier/hooks/runs/1935645842?check_suite_focus=true
My intent is to have an idea of the breakage this can cause in existing packages. I'm pretty sure that github won't allow me to inspect the log in the end though, it'll probably be too big.

ericLemanissier:
happy surprise, only 4 recipes trigger an error with this hook, so I submitted the fixes:

1. https://github.com/conan-io/conan-center-index/pull/4615
2. https://github.com/conan-io/conan-center-index/pull/4614
3. https://github.com/conan-io/conan-center-index/pull/4613
4. https://github.com/conan-io/conan-center-index/pull/4610

prince-chrismc (approved):

uilianries (approved):

danimtb (approved):

SpaceIm:
I don't know if it comes from this hook, but now all test_package fail with undefined references.

hooks/conan-center.py
```diff
@@ -841,6 +842,27 @@ def test(out):
             out.error("Found files: {}".format("; ".join(files_missplaced)))
 
 
+    @run_test("KB-H054", output)
+    def test(out):
+        def _test_component(component):
+            libs_to_search = component.libs
```
SpaceIm, line 848: reference of component.libs !

hooks/conan-center.py
```diff
@@ -841,6 +842,27 @@ def test(out):
...
+            for p in component.libdirs:
+                libs_found = tools.collect_libs(conanfile, p)
+                if not libs_found:
+                    out.warn("Component %s::%s libdir \"%s\" does not contain any library" % (conanfile.name, component.name, p))
+                libs_declared_and_found = [l for l in libs_found if l in libs_to_search]
+                if not libs_declared_and_found:
+                    out.warn("Component %s::%s libdir \"%s\" does not contain any declared library" % (conanfile.name, component.name, p))
+                for l in libs_declared_and_found:
+                    libs_to_search.remove(l)
```
SpaceIm, line 857: and here it removes all libs ! kaboom !

uilianries: ahaha true! I was about to comment it now. We have to copy instead. But first, I'll provide a test for this hook. Thanks

SpaceIm:
> * if some libdir contain libraries, but none of these are declared in the libs attribute (warning)

Do we want to display a warning for this? Plugins are not listed in cpp_info.libs.

uilianries:
> So we'll see warnings in all header-only libraries, and have to fix all of them (self.cpp_info.libdirs = []) to reduce verbosity.

Indeed, it can be improved in a next PR. Let's fix this little bug first.

jgsogo:
Now that it is working again thanks to #277, time to come back to the implementation. I think that the two comments by @SpaceIm make a lot of sense. It is Conan the one adding some default path, we cannot generate warnings for those defaults...

PR #277, opened 2021-02-25
[hotfix] KB-H054 must not modify cpp_info

uilianries:
After KB-H054 we started to have a lot of trouble in CCI, it seems like we fall for the Python prank :sweat_smile: 

Related to #273

Thanks to @SpaceIm for pointing us that bug! 

- Copy cpp_info.libs instead of using its reference
- Add new tests for KB-H054

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -749,7 +749,7 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
     def test(out):
         if conanfile.version == "system":
             return
-        
+
         # INFO: Whitelist for package names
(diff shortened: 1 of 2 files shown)


danimtb (approved):

uilianries:
> Sorry about that :/

No way! We reviewed and approved too. It's not your fault! I committed the same error in the past, that's what happens when a C++ developer touches Python ;)

jgsogo (approved):

SSE4 (approved):

jgsogo:
West or east, where is the `const` in Python? 😑

PR #278, opened 2021-02-26
KB-H054 should filter header-only recipes

uilianries:
Related to https://github.com/conan-io/hooks/pull/273#issuecomment-786257908

We don't need to check header-only when libdir is empty.

/cc @SpaceIm


ericLemanissier:
I'd prefer the opposite: make conan set the default cpp_info.lib_paths empty if the settings do not have os compiler and arch

SpaceIm:
What about tools only recipes?

ericLemanissier:
it would be better if they set `self.cpp_info.libdirs` to an empty list right ? To me this message is a real information, not spammy message.

uilianries:
> I'd prefer the opposite: make conan set the default cpp_info.lib_paths empty if the settings do not have os compiler and arch

Yeah, but it would be a feature for Conan. Conan Center hook shouldn't change the recipe.

ericLemanissier:
Yes, this is what I have in mind

SpaceIm:
It should also filter the global target of recipes with components. Currently it displays a warning for all recipes with components .

jgsogo (approved):
At least, we prevent the error for header-only libs. We need to have a look at the components too.

danimtb (approved):

Issue #279, opened 2021-03-03
conan-center hooks 'final report'


ericriff:
Having the conan-center hook activated is super useful to find common errors / tidy up your code, but under some circumstances it is extremely chatty which makes easy to miss an error / warning since they fly by so quickly and get buried under a million lines of log.

I don't know if this is possible but I would like to have a 'final report' in my console after a conan create execution that tells me if the hook detected errors or not (and even better list the errors).

Another thing would be to make the hook display only warnings and errors, and not the 'OK' lines.

Thanks!

uilianries:
Conan-center hooks uses `output.success` which a regular message output. Conan uses it a lot, se we can't set a global log level. What we could is adding an env var to filter it.

uilianries:
It can be related to https://github.com/conan-io/conan/issues/8570. We provide a final report, we would need a post command hook. For now, we can't provide a final report, as depending which conan command we run, it will or not execute a step.

What we do now, is silencing INFO and printing only WARNING and ERROR.

PR #280, opened 2021-03-11
Add logging level for conan-center hook

uilianries:
This PR is an extension of `CONAN_LOGGING_LEVEL`, but for Conan Center hook.

For local development we may not want to see all info, only warning and error are enough.

It's related to #279. Unfortunately we don't have a hook post-command yet, but we can silence minor messages at least.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -2,7 +2,7 @@ import fnmatch
 import inspect
 import os
 import re
-from logging import WARNING, ERROR, INFO, DEBUG, NOTSET
+from logging import WARNING, ERROR, INFO, DEBUG, NOTSET, _nameToLevel
 
 import yaml
 from conans import tools, Settings
@@ -67,6 +67,11 @@ class _HooksOutputErrorCollector(object):
         if self.kb_id:
             self.kb_url = kb_url(self.kb_id)
         self._error_level = int(os.getenv("CONAN_HOOK_ERROR_LEVEL", str(NOTSET)))
+        self._set_logging_level()
+
+    def _set_logging_level(self):
+        level = os.getenv("CONAN_HOOK_LOGGING_LEVEL", str(NOTSET))
+        self._logging_level = int(level) if level.isdigit() else _nameToLevel[level.upper()]
 
     def _get_message(self, message):
         if self._test_name:
@@ -76,22 +81,26 @@ class _HooksOutputErrorCollector(object):
             return message
 
     def success(self, message):
-        self._output.success(self._get_message(message))
+        if self._logging_level < WARNING:
+            self._output.success(self._get_message(message))
 
     def debug(self, message):
         if self._error_level and self._error_level <= DEBUG:
             self._error = True
-        self._output.debug(self._get_message(message))
+        if self._logging_level < INFO:
+            self._output.debug(self._get_message(message))
 
     def info(self, message):
         if self._error_level and self._error_level <= INFO:
             self._error = True
-        self._output.info(self._get_message(message))
+        if self._logging_level < WARNING:
+            self._output.info(self._get_message(message))
 
     def warn(self, message):
         if self._error_level and self._error_level <= WARNING:
             self._error = True
-        self._output.warn(self._get_message(message))
+        if self._logging_level < ERROR:
+            self._output.warn(self._get_message(message))
 
     def error(self, message):
         self._error = True
--- a/tests/test_hooks/conan-center/test_conan-center.py
+++ b/tests/test_hooks/conan-center/test_conan-center.py
@@ -1035,3 +1035,41 @@ class ConanCenterTests(ConanClientTestCase):
         output = self.conan(['create', '.', 'name/version@user/test'])
         self.assertIn("[LIBRARY DOES NOT EXIST (KB-H054)] OK", output)
         self.assertNotIn('does not contain any library', output)
+
+    def test_logging_level(self):
+        conanfile = textwrap.dedent("""\
+        from conans import ConanFile
+        import os
+
+        class FoobarConan(ConanFile):
+            author = "Foobar"
+
+            def package(self):
+                pass
+        """)
+        tools.save('conanfile.py', content=conanfile)
+        output = self.conan(['create', '.', 'name/version@user/test'])
+        self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
+        self.assertIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
+        self.assertIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
+
+        for level in ["INFO", "20"]:
+            with tools.environment_append({"CONAN_HOOK_LOGGING_LEVEL": level}):
+                output = self.conan(['create', '.', 'name/version@user/test'])
+                self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
+                self.assertIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
+                self.assertIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
+
+        for level in ["Warning", "30"]:
+            with tools.environment_append({"CONAN_HOOK_LOGGING_LEVEL": level}):
+                output = self.conan(['create', '.', 'name/version@user/test'])
+                self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
+                self.assertIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
+                self.assertNotIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
+
+        for level in ["error", "40"]:
+            with tools.environment_append({"CONAN_HOOK_LOGGING_LEVEL": level}):
+                output = self.conan(['create', '.', 'name/version@user/test'])
+                self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
+                self.assertNotIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
+                self.assertNotIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
--- a/README.md
+++ b/README.md
@@ -88,7 +88,16 @@ variable ``CONAN_HOOK_ERROR_LEVEL``:
    - ``CONAN_HOOK_ERROR_LEVEL=40`` it will raise if any error happen.
    - ``CONAN_HOOK_ERROR_LEVEL=30`` it will raise if any error or warning happen.
 
+If you want to decrease logging level, you can adjust it by the environment variable
+``CONAN_HOOK_LOGGING_LEVEL``:
 
+  - ``CONAN_HOOK_LOGGING_LEVEL=40`` it will print out only error messages
+  - ``CONAN_HOOK_LOGGING_LEVEL=WARNING`` it will print out only warning and error messages
+
+It accepts both logging level names and logging level code:
+  - ERROR: 40
+  - WARNING/WARN: 30
+  - INFO: 20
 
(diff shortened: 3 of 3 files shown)


jgsogo (commented):
Do we want to name the variable `CONAN_HOOK...` when it is something that affects only to the hook named `conan-center`?

uilianries:
Well, we already have `CONAN_HOOK_ERROR_LEVEL`. But we can re-use them for other hooks too, it's just because we don't have a separated feature it on Conan. I just want to keep it separated from `CONAN_LOGGING_LEVEL`. Because if I work developing recipes for my company and cci at same machine, I'll want to mute conan-center hook messages, but keeping regular Conan client messages working.

jgsogo (dismissed):
Yeah, the same pattern as `CONAN_HOOK_ERROR_LEVEL`

hooks/conan-center.py
```diff
@@ -76,22 +82,26 @@ def _get_message(self, message):
...
     def success(self, message):
-        self._output.success(self._get_message(message))
+        if self._logging_level < WARNING:
+            self._output.success(self._get_message(message))
 
     def debug(self, message):
         if self._error_level and self._error_level <= DEBUG:
             self._error = True
-        self._output.debug(self._get_message(message))
+        if self._logging_level < INFO:
```
SSE4, line 91: probably it should be `if self._logging_level <= WARNING:` ?

I actually don't like the code need to check level before the output.
I guess these functions (info/warning/error) should check the level itself, and render no-op if level is too high

uilianries: > probably it should be if self._logging_level <= WARNING: ?

Not for debug messages. It's lower level than INFO:

https://docs.python.org/3/library/logging.html#levels

> I actually don't like the code need to check level before the output.

Me too, but it's a logger from Conan client, we would need initialize a new one only for hooks, but also I dunno if it's a singleton.

jgsogo:
In C3i we should probably use WARNING level, right? That way when we tell a user that the recipe is broken because of the hook, they will see only the warnings/errors, but not all the hooks that have passed. Am I right?

uilianries:
@jgsogo Yes, in CCI it would be nice using WARNING as default level. I don't know if we should enable it by default or CCI should add CONAN_HOOK_LOGGING_LEVEL

hooks/conan-center.py
```diff
@@ -67,6 +67,12 @@ def __init__(self, output, kb_id=None):
         if self.kb_id:
             self.kb_url = kb_url(self.kb_id)
         self._error_level = int(os.getenv("CONAN_HOOK_ERROR_LEVEL", str(NOTSET)))
+        self._set_logging_level()
+
+    def _set_logging_level(self):
+        level = os.getenv("CONAN_HOOK_LOGGING_LEVEL", str(NOTSET))
+        name_level = {"ERROR": 40, "WARNING": 30, "WARN": 30, "INFO": 20, "DEBUG": 10, "NOTSET": 0}
+        self._logging_level = int(level) if level.isdigit() else name_level[level.upper()]
```
madebr, line 75: What if level is not in the dict?
I propose to output an error and set the warning level to `NOTSET`.
```suggestion
        if level.isdigit():
            self._logging_level = int(level)
        elif level in name_level:
            self._logging_level = name_level[level.upper()]
        else:
            self._logging_level = NOTSET
            self.error("CONAN_HOOK_LOGGING_LEVEL is set to an incorrect value")
```

hooks/conan-center.py
```diff
@@ -76,22 +82,26 @@ def _get_message(self, message):
             return message
 
     def success(self, message):
-        self._output.success(self._get_message(message))
+        if self._logging_level < WARNING:
```
madebr, line 85: Do we care about success messages?
```suggestion
        if self._logging_level <= INFO:
```

hooks/conan-center.py
madebr: I'm thinking this should be:
```suggestion
        if self._logging_level <= DEBUG:
```

hooks/conan-center.py
```diff
@@ -76,22 +82,26 @@ def _get_message(self, message):
...
             self._error = True
-        self._output.debug(self._get_message(message))
+        if self._logging_level < INFO:
+            self._output.debug(self._get_message(message))
 
     def info(self, message):
         if self._error_level and self._error_level <= INFO:
             self._error = True
-        self._output.info(self._get_message(message))
+        if self._logging_level < WARNING:
```
madebr, line 97:
```suggestion
        if self._logging_level <= INFO:
```

hooks/conan-center.py
```diff
@@ -76,22 +82,26 @@ def _get_message(self, message):
...
+            self._output.info(self._get_message(message))
 
     def warn(self, message):
         if self._error_level and self._error_level <= WARNING:
             self._error = True
-        self._output.warn(self._get_message(message))
+        if self._logging_level < ERROR:
```
madebr, line 103:
```suggestion
        if self._logging_level <= WARNING:
```

hooks/conan-center.py
```diff
@@ -76,22 +82,26 @@ def _get_message(self, message):
...
+            self._output.warn(self._get_message(message))
 
     def error(self, message):
         self._error = True
```
madebr, line 107: I would also add a check to this method.
In theory, you can set the verbosity level very high.
```python
if self._logging_level <= ERROR:
    ...
```

Croydon:
I would like WARNING level per default locally too. I think that at this point all the `OK`s are kinda spam in the logs 🙈

hooks/conan-center.py
```diff
@@ -77,6 +77,18 @@ def __init__(self, output, kb_id=None):
...
             self.kb_url = kb_url(self.kb_id)
         self._error_level = int(os.getenv("CONAN_HOOK_ERROR_LEVEL", str(NOTSET)))
+        self._set_logging_level()
+
+    def _set_logging_level(self):
+        level = os.getenv("CONAN_HOOK_LOGGING_LEVEL", str(NOTSET))
+        name_level = {"ERROR": 40, "WARNING": 30, "WARN": 30, "INFO": 20, "DEBUG": 10, "NOTSET": 0}
+        if level.isdigit():
+            self._logging_level = int(level)
+        elif level in name_level:
```
madebr, line 87:
```suggestion
        elif level.UPPER() in name_level:
```

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -1057,6 +1057,45 @@ def package_info(self):
...
+            def package(self):
+                pass
+        """)
+        tools.save('conanfile.py', content=conanfile)
+        output = self.conan(['create', '.', 'name/version@user/test'])
+        self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
+        self.assertIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
+        self.assertIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
+
+        for level in ["INFO", "20"]:
```
madebr, line 1079:
```suggestion
        self.assertIn("[FPIC MANAGEMENT (KB-H007)] OK", output)

        with tools.environment_append({"CONAN_HOOK_LOGGING_LEVEL": "oops"}):
           output = self.conan(['create', '.', 'name/version@user/test'])
           self.assertIn("ERROR: CONAN_HOOK_LOGGING_LEVEL is set to an incorrect value", output)

        for level in ["INFO", "20"]:
```

hooks/conan-center.py
uilianries: Done

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -1057,6 +1057,45 @@ def package_info(self):
...
+    def test_logging_level(self):
+        conanfile = textwrap.dedent("""\
+        from conans import ConanFile
+        import os
+
+        class FoobarConan(ConanFile):
+            author = "Foobar"
+
+            def package(self):
+                pass
```
madebr, line 1071:
```suggestion
        conanfile = textwrap.dedent("""\
        from conans import ConanFile

        class FoobarConan(ConanFile):
            pass
```

tests/test_hooks/conan-center/test_conan-center.py
```diff
@@ -1057,6 +1057,45 @@ def package_info(self):
...
+                self.assertNotIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
+
+        for level in ["error", "40"]:
+            with tools.environment_append({"CONAN_HOOK_LOGGING_LEVEL": level}):
+                output = self.conan(['create', '.', 'name/version@user/test'])
+                self.assertIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
+                self.assertIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
+                self.assertNotIn("[FPIC MANAGEMENT (KB-H007)] OK", output)
 
     def test_os_rename_warning(self):
```
madebr, line 1100:
```suggestion
                self.assertNotIn("[FPIC MANAGEMENT (KB-H007)] OK", output)

        with tools.environment_append({"CONAN_HOOK_LOGGING_LEVEL": "9001"}):  # Over 9000
            output = self.conan(['create', '.', 'name/version@user/test'])
            self.assertNotIn("ERROR: [PACKAGE LICENSE (KB-H012)]", output)
            self.assertNotIn("WARN: [HEADER_ONLY, NO COPY SOURCE (KB-H005)]", output)
            self.assertNotIn("[FPIC MANAGEMENT (KB-H007)] OK", output)

    def test_os_rename_warning(self):
```

hooks/conan-center.py
uilianries: good catch

tests/test_hooks/conan-center/test_conan-center.py
uilianries: Over_9000!

tests/requirements_test.txt
```diff
@@ -2,6 +2,6 @@ pytest>=3.6
 parameterized
 responses
 pluggy==0.11.0
-pylint
+pylint==2.10.2
```
uilianries, line 5: Pylint 2.11 is causing errors related to improper objects (expected bytes but they changed to string). If you check pylint issues, there are more complains about that version

uilianries:
I needed to force Pylint 2.10, because 2.11 changed something which broke hooks. I did a quick read on Pylint repo and there are more issues related to 2.11. My suggestion is using 2.10 for now and wait for a solution (maybe 2.12).

madebr:
About logging levels, I think we want 2 things.
1. less verbose logs when building locally, which can be accomplished by this pr
2. increase the severity of an error (aka introduce `-Werror`):
I think it would be useful for c3i to mark warnings as errors.
Right now, we have some checks marked as warnings because marking them as error would cause too many builds to fail.
To that, I propose to only increase the severity if the name of the recipe is the same as the recipe is being built (and not a dependency).

uilianries:
>  I think it would be useful for c3i to mark warnings as errors.

It will break many recipes in Conan Center. You are a Conan expert, know about the errors and warnings, but for new contributors, it can be a block when submitting their first PR.

madebr:
It's a double edged sword.
I think we should think about ways to give attention to the warnings then.
Right now, c3i hides them.
It would be useful to have them available somewhere such that people/reviewers can propose fixes for them.

Croydon:
I would like to see that the CCI bot is outputting all hook warnings as a comment in the PR, but not blocking the PR because of it

SSE4:
> I would like to see that the CCI bot is outputting all hook warnings as a comment in the PR, but not blocking the PR because of it

yes, it's possible, we were thinking how to implement it in CI.
but as we run initial export + parallel builds, it has to be gathered in multiple places and then somehow represented in the readable way, which is not too annoying.

Croydon:
What is the current status of this?

uilianries:
Need to re-visit

PR #281, opened 2021-03-12
Fix KB-H054 for header-only case

uilianries:
One missing case for this hook.

Related cases:

https://github.com/conan-io/conan-center-index/pull/4856/files#r593086602
https://github.com/conan-io/hooks/pull/278

UPDATE:

I removed all warning messages, because header-only is only the top of iceberg, we also need to filter deps components and local components, look what is consumed and what's not. For now we have false-positives which are not cool.

The hook 054 works good for its task, but I don't think those extra warning are relevant now.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -863,7 +863,7 @@ def post_package_info(output, conanfile, reference, **kwargs):
                 if not libs_found and not _is_recipe_header_only(conanfile):
                     out.warn("Component %s::%s libdir \"%s\" does not contain any library" % (conanfile.name, component.name, p))
                 libs_declared_and_found = [l for l in libs_found if l in libs_to_search]
-                if not libs_declared_and_found:
+                if not libs_declared_and_found and not _is_recipe_header_only(conanfile):
                     out.warn("Component %s::%s libdir \"%s\" does not contain any declared library" % (conanfile.name, component.name, p))
                 for l in libs_declared_and_found:
                     libs_to_search.remove(l)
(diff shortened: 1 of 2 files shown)


SSE4 (approved):

danimtb (approved):

danimtb:
@uilianries please check the CI

hooks/conan-center.py
```diff
@@ -859,12 +859,6 @@ def test(out):
         def _test_component(component):
             libs_to_search = list(component.libs)
             for p in component.libdirs:
-                libs_found = tools.collect_libs(conanfile, p)
-                if not libs_found and not _is_recipe_header_only(conanfile):
-                    out.warn("Component %s::%s libdir \"%s\" does not contain any library" % (conanfile.name, component.name, p))
-                libs_declared_and_found = [l for l in libs_found if l in libs_to_search]
-                if not libs_declared_and_found:
-                    out.warn("Component %s::%s libdir \"%s\" does not contain any declared library" % (conanfile.name, component.name, p))
```
ericLemanissier, line 867: this should be 
```
                libs_found = tools.collect_libs(conanfile, p)
                libs_declared_and_found = [l for l in libs_found if l in libs_to_search]
```
because `libs_found` and `libs_declared_and_found` are used afterwards

uilianries: Thank you! I've added!

jgsogo (approved):

jgsogo:
This is safe to merge, right? We are just removing extra warnings... it cannot break anything 😄

PR #283, opened 2021-03-16
Name attribute is mandatory

uilianries:
Unfortunately `getattr` doesn't work, because Conan injects `name` from package reference. We have to check directly in file content. 

closes #282

Issue #282, opened 2021-03-13
[conan center] check name attribute


SSE4:
I initially forgot the `name` attribute in https://github.com/conan-io/conan-center-index/pull/4856, but CI didn't complain

PR #284, opened 2021-03-17
Accepts tabs and multiple spaces for name attr

uilianries:
Related to https://github.com/conan-io/hooks/pull/283 and #282

As commented by @ericLemanissier and @SSE4, #282 only covers basic cases, but a recipe can use different combinations of spaces and tabs.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -174,7 +174,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                 if not field_value:
                     out_method("Conanfile doesn't have '%s' attribute. " % field)
 
-        if "    name =" not in conanfile_content:
+        if not re.search(r"(\s{4}|\t)name\s*=", conanfile_content):
             out.error("Conanfile doesn't have 'name' attribute.")
         _message_attr(["url", "license", "description", "homepage", "topics"], out.error)
 
--- a/tests/test_hooks/conan-center/test_conan-center.py
+++ b/tests/test_hooks/conan-center/test_conan-center.py
@@ -450,6 +450,15 @@ class ConanCenterTests(ConanClientTestCase):
             self.assertNotIn(msg, output)
         self.assertIn("[RECIPE METADATA (KB-H003)] OK", output)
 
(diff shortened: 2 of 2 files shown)


hooks/conan-center.py
```diff
@@ -174,7 +174,7 @@ def _message_attr(attributes, out_method):
                 if not field_value:
                     out_method("Conanfile doesn't have '%s' attribute. " % field)
 
-        if "    name =" not in conanfile_content:
+        if not re.search(r"(\s{4}|\t)name\s*=", conanfile_content):
```
SSE4, line 177: is there regex which allows any space characters? https://en.wikipedia.org/wiki/Whitespace_character#Unicode

jgsogo: I think this is what we want: four spaces or a single tab.

jgsogo (approved):
I understand the concern, but I'm not sure if this is needed. I run the https://github.com/conan-io/conan-center-index/issues/2232 just after merging (just in case 😅  ) and all the recipes work with this new hook.... so all of them are using spaces instead of tabs.

Nevertheless, this is not the hook that should check spaces/tabs, so it is ok to make it _generic_

hooks/conan-center.py
uilianries: Legit, but not used (fortunately).

SSE4 (approved):

hooks/conan-center.py
SSE4: if no other characters are used right now, I am okay

jgsogo: Yes, that's the point. It is legit but we don't want it in `conan-center-index`.

My main concern is if there were any usage of `set_name(self)` for any reason... but I think it is something we won't find in `conan-center-index`.... and now this hook will fail and we will get a notification from the user.

SSE4: I guess `set_name` could be used, e.g. if I want to read directory name and use it as a name, it will allow templates to be easier - you don't need to edit name after `conan new`... but I am not sure I like it

danimtb (approved):

PR #286, opened 2021-03-22
Remove fPIC when building shared

### LINKED ISSUE #285: [conan-center] fPIC should not affect package id when building shared library
Hi!

Using `fPIC` option with `shared` is kind of mandatory, so we remove `fPIC` when building `shared` library. It avoids illegal package ID, as that option does not allow an invalid package (shared + non fPIC).

That rule is very used on CCI recipes, indeed more than 300 recipes there follows such rule:

    pcregrep -r -M 'if self.options.shared:\n\s+del.self.options.fPIC' recipes/ | wc -l

It would be great having it as default check for CCI recipes.

uilianries:
It's recommendation well followed. More than 300 recipes in CCI are using it.

Wiki: https://github.com/conan-io/conan-center-index/pull/5002

closes #285

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -718,7 +718,7 @@ def pre_build(output, conanfile, **kwargs):
             out.error("'fPIC' option not managed correctly. Please remove it for Windows "
                       "configurations: del self.options.fpic")
             error = True
-        if conanfile.options.get_safe("fPIC") and conanfile.options.get_safe("shared"):
+        if has_fpic and conanfile.options.get_safe("shared"):
             out.error("'fPIC' option not managed correctly. Please remove it for shared "
                       "option: del self.options.fpic")
             error = True
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -713,10 +713,16 @@ def pre_build(output, conanfile, **kwargs):
     @run_test("KB-H007", output)
     def test(out):
         has_fpic = conanfile.options.get_safe("fPIC")
+        error = False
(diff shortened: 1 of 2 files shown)


SSE4:
I am okay if it's needed, but keep in mind hook may fail for many existing recipes

SSE4 (approved):

jgsogo (approved):
... and we cannot test this hook in advance to know how many will fail :/

ericLemanissier:
it could be done with a github action (for example https://github.com/ericLemanissier/hooks/blob/main/main.py#L84)

hooks/conan-center.py
```diff
@@ -713,10 +713,16 @@ def pre_build(output, conanfile, **kwargs):
     @run_test("KB-H007", output)
     def test(out):
         has_fpic = conanfile.options.get_safe("fPIC")
+        error = False
         if conanfile.settings.get_safe("os") == "Windows" and has_fpic:
             out.error("'fPIC' option not managed correctly. Please remove it for Windows "
                       "configurations: del self.options.fpic")
-        elif has_fpic:
+            error = True
+        if conanfile.options.get_safe("fPIC") and conanfile.options.get_safe("shared"):
```
danimtb, line 721:
```suggestion
        if has_fpic and conanfile.options.get_safe("shared"):
```

danimtb (approved):

danimtb:
> it could be done with a github action (for example https://github.com/ericLemanissier/hooks/blob/main/main.py#L84)

Indeed, that would be a nice addition to the conan-center hook CI

PR #294, opened 2021-04-28
KB-H057 - os.rename is not recommended

uilianries:
The `os.rename` is no longer recommended, instead, use `tools.rename`

This PR adds this rule as warning, otherwise CCI will break heavily.

Related to #293


uilianries:
> ERROR: Could not install requirement enum34>=1.1.3; python_version < "3.4" from https://files.pythonhosted.org/packages/6f/2c/a9386903ece2ea85e9807e0e062174dc26fdce8b05f216d00491be29fad5/enum34-1.1.10-py2-none-any.whl#sha256=a98a201d6de3f2ab3db284e70a33b0f896fbf35f8086594e8c9e74b909058d53 (from astroid->-r C:\projects\hooks/tests/requirements_test.txt (line 6)) because of HTTP error 502 Server Error: Gateway Error for url: https://files.pythonhosted.org/packages/6f/2c/a9386903ece2ea85e9807e0e062174dc26fdce8b05f216d00491be29fad5/enum34-1.1.10-py2-none-any.whl for URL https://files.pythonhosted.org/packages/6f/2c/a9386903ece2ea85e9807e0e062174dc26fdce8b05f216d00491be29fad5/enum34-1.1.10-py2-none-any.whl#sha256=a98a201d6de3f2ab3db284e70a33b0f896fbf35f8086594e8c9e74b909058d53 (from https://pypi.org/simple/enum34/)

jgsogo:
Right now this is probably the best approach to implement it... but in order to merge it and make it useful, we need to add some functionality to the library https://github.com/conan-io/c3i_jenkins/issues/758. 

Let's wait a little bit, see how we implement it in the library before merging this one, maybe we come up with some changes to improve it.

jgsogo (changes requested):
We will suggest user to use the new (to come) `rename` function that will let us configure the usage of  `robocopy`: https://github.com/conan-io/conan/issues/8897

We will update this PR once the functionality is ready to be used.

uilianries:
Done, now the recommended way is `tools.rename(self, src, dst)`

Croydon:
`tools.rename(self, src, dst)` does not seem to be documented right now. Also why does it need the self reference?

uilianries:
@Croydon It's documented here: https://docs.conan.io/en/latest/reference/conanfile/tools/files.html#conan-tools-rename

The feature issue is this one: https://github.com/conan-io/conan/issues/8897
But the origin of that topic is this one: https://github.com/conan-io/conan/issues/7105

The `self` is part of remodeling tools module. It has been worked as 'utils', where we putted anything there and consumed as external package. Now, accessing `self`, we will keep the same pattern for for those methods. For instance, cross-building method consumes settings, but now we can use self, instead.

jgsogo:
I would add the full import chain to the error message: _"... use `conan.tools.rename(self, ....`"_ otherwise people might not be aware of the new tool. It's not just about adding `self` as the first argument.

Croydon:
@jgsogo This was my confusion as I looked in the wrong place in the documentation

prince-chrismc (approved):

danimtb (dismissed):

danimtb:
We need Conan 1.37.0 in the CCI pipelines at least before merging this

SSE4:
1.37 is already in CCI, right?

SSE4 (approved):

uilianries:
Yes, 1.37.2 is running on CCI right now! /cc @danimtb

jgsogo (approved):
Time to approve this one 🚀

Issue #301, opened 2021-06-02
Broken OSX Travis CI jobs


uilianries:
The PR #300 failed due a suspicious error involving Travis and its OSX jobs.

https://travis-ci.org/github/conan-io/hooks/builds/773294565

I deserves a further investigation.

/cc @jgsogo

Croydon:
Unless the Conan team is willing to regularly request new credits from Travis, it will stop running eventually anyway. 

See https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

Maybe migrate to GitHub Actions?

uilianries:
That's a possibility which I would vote, but probably team will vote for Jenkins, because is under their control.

jgsogo:
Yes, we need to migrate from Travis/Appveyor. Github-Actions is ok, our Jenkins is another alternative. This is a very small project, so I think we shouldn't consider any other factors.

Croydon:
This can be closed

PR #304, opened 2021-06-07
Move CI to Jenkins

danimtb:
- Added Jenkinsfile with declarative syntax
- Created scripts to handle the environment setup for each platform
- Only testing in Windows and Linux at the moment (we can add macos later).
- Reduced the combination of pythons x conans to avoid having a slow CI.
- Note that the travis and appveyor files are still not removed (will be done later when Jenkins is configured).

Tested the branch here: https://ci.conan.io/blue/organizations/jenkins/Hooks/detail/feature%2Fmove_ci_to_jenkins/38/pipeline
image

Related to #301
closes #258


.ci/Jenkinsfile
```diff
@@ -0,0 +1,180 @@
+pipeline {
+    agent none
+
+    stages {
+        stage("Windows") {
```
danimtb, line 5: In windows we don't have to install the python interpreter, so the setup is faster, so they are run first

.ci/Jenkinsfile
```diff
@@ -0,0 +1,180 @@
...
+                            }
+                        }
+                        stage('Linux - py35 - conandev') {
+                            environment {
+                                TOXENV = 'py35-conandev'
+                            }
+                            steps {
+                                sh '''
+                                    eval "$(pyenv init -)"
+                                    pyenv activate conan
```
danimtb, line 95: Didn't want to add anything related to the environment at this step (only run tests), but I couldn't figure it out so I had to place this two lines here

danimtb:
Linux builds can be improved as some python versions are already installed in the images (see https://github.com/conan-io/conan/blob/develop2/.ci/jenkins/testsv2.jenkins), but I think it is fine ATM

Croydon:
Will the logs be still publicly visible for anyone?

danimtb:
@Croydon yes, they should be available in the same way the conan-io/conan ones are right now

jgsogo:
Just checking notifications 😉

jgsogo (approved):
Good!

uilianries (approved):
I'm not well familiarized with jenkinsfile, but it looks well ported here. As we are using tox, we don't need to add more steps. This PR is a really great addition! I would a status badge on README file if possible, so we can check the master branch easily.

uilianries:
BTW, can we validate it on OSX too? It would be great, we usually don't have problems in Mac, but it would be a good test scenario too.

danimtb:
> BTW, can we validate it on OSX too? It would be great, we usually don't have problems in Mac, but it would be a good test scenario too.

yes, we can add some Macos configurations as well in a next PR as well as the badge in the readme. I will take a look!

Issue #298, opened 2021-05-21
Disallow illegal characters in filenames


madebr:
Windows puts more limitations on what characters a valid filename can contain.
Using an invalid character, e.g. a `:`, causes weird problems in c3i.
See https://github.com/conan-io/conan-center-index/pull/5583

Maybe it would be interesting to create a hook (or amend a hook) to check whether all filenames are valid?

Windows naming conventions:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

uilianries:
Move to Linux, problem solved 😈 🐧 

Besides the joke, it's a valid case indeed, thanks for suggesting.

PR #311, opened 2021-06-14
KB-H058 Disallowed characters

uilianries:
Hello! This PR implements #298 

Any reserved character won't be allowed anymore. The error message will list the file name and ask for renaming.

closes https://github.com/conan-io/hooks/issues/298

/cc @madebr

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -612,13 +612,16 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H058", output)
     def test(out):
-        disallowed_chars = '<>:"/\\|?*'
+        disallowed_chars = '<>:"/\\|?*%,; '
         recipe_folder = os.path.dirname(conanfile_path)
         for root, _, files in os.walk(recipe_folder):
             for file in files:
                 if any(it in disallowed_chars for it in file):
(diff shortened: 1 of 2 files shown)


madebr (approved):
Looks good to me!

hooks/conan-center.py
```diff
@@ -609,6 +610,16 @@ def test(out):
                 out.error("Both '{0}requires' attribute and '{0}requirements()' method should not "
                           "be declared at same recipe.".format(prefix))
 
+    @run_test("KB-H058", output)
+    def test(out):
+        disallowed_chars = '<>:"/\\|?*'
```
jgsogo, line 615: Even if they are not strictly forbidden I would add some more rules in this hook. Most taken from here (https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words):
 * characters `,`, `;`, `%`
 * blank/spaces
 * `.` at the end

Can you figure out any reason to have those characters in file/folder names?

jgsogo (commented):
I like it, preventing weird filenames is a condition to sanity 😄

hooks/conan-center.py
uilianries: patch files and extra testing files (e.g xml), people can use any name they "want".

jgsogo (approved):

madebr (approved):

PR #317, opened 2021-07-01
KB-H060 Read only gitattribute

uilianries:
Reading .gitattibute file would be a problem, as we only have the recipe folder path. It could use relative path, or even a git python library to load it, but I think is too much.

Let's use the common extensions listed in .gitattribute file, it should cover 99% of our cases.

I also added .json .xml in case some exotic project need an extra configuration file(Boost and Qt for instance) or even for some json parser which uses a json file in test_package

/cc @SpaceIm

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -634,13 +634,16 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H060", output)
     def test(out):
-        allowed_ext = [".cmake", ".conf", ".cfg", ".diff", ".md", ".patch", ".py", ".txt", ".yml", ".am", ".xml", ".json", ".in", ".ac", ".tsx", ".tmx", ".proto", ".capnp", ".c", ".cc", ".c++", ".cpp", ".cxx", ".c++m", ".cppm", ".cxxm", ".h++", ".hh", ".hxx", ".hpp"]
+        ext_to_be_checked = [".cmake", ".conf", ".cfg", ".diff", ".md", ".patch", ".py", ".txt",
+                             ".yml", ".am", ".xml", ".json", ".in", ".ac", ".tsx", ".tmx"
+                             ".proto", ".capnp", ".c", ".cc", ".c++", ".cpp", ".cxx", ".c++m",
+                             ".cppm", ".cxxm", ".h++", ".hh", ".hxx", ".hpp", ".qrc", ".pro", ".build "]
         recipe_folder = os.path.dirname(conanfile_path)
         for root, _, files in os.walk(recipe_folder):
             if os.path.relpath(root, recipe_folder).replace("\\", "/").startswith("test_package/build"):
                 continue
             for filename in files:
-                if not any(filename.endswith(ext) for ext in allowed_ext):
+                if not any(filename.endswith(ext) for ext in ext_to_be_checked):
                     continue
                 lines = open(os.path.join(root, filename), 'rb').readlines()
                 if any(line.endswith(b'\r\n') for line in lines):
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -634,7 +634,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H060", output)
     def test(out):
-        allowed_ext = [".cmake", ".conf", ".diff", ".md", ".patch", ".py", ".txt", ".yml", ".am", ".xml", ".json"]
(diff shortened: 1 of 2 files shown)


jgsogo (approved):

danimtb (approved):

ericLemanissier:
Wouldn't it be better to do the test as post_export, and check only what is in the local cache ?

hooks/conan-center.py
```diff
@@ -634,11 +634,14 @@ def test(out):
 
     @run_test("KB-H060", output)
     def test(out):
+        allowed_ext = [".cmake", ".conf", ".diff", ".md", ".patch", ".py", ".txt", ".yml", ".am", ".xml", ".json"]
```
SpaceIm, line 637:
```suggestion
        allowed_ext = [".cmake", ".conf", ".diff", ".md", ".patch", ".py", ".txt", ".yml", ".am", ".xml", ".json", ".c", ".cc", ".c++", ".cpp", ".cxx", ".c++m", ".cppm", ".cxxm", ".h++", ".hh", ".hxx", ".hpp"]
```

SpaceIm:
Maybe also .qrc and .pro for qmake, and .build for meson

uilianries:
> Wouldn't it be better to do the test as post_export, and check only what is in the local cache ?

local cache doesn't contain test_package

> EDIT: allowing .patch and .diff files to have CRLF is a big hole in the net, because a lot of recipe exports at least one of these

We don't allow, maybe that variable name is bad, I'll update it.

hooks/conan-center.py
```diff
@@ -634,11 +634,17 @@ def test(out):
 
     @run_test("KB-H060", output)
     def test(out):
+        ext_to_be_checked = [".cmake", ".conf", ".cfg", ".diff", ".md", ".patch", ".py", ".txt",
+                             ".yml", ".am", ".xml", ".json", ".in", ".ac", ".tsx", ".tmx"
```
SpaceIm, line 638:
```suggestion
                             ".yml", ".am", ".xml", ".json", ".in", ".ac", ".tsx", ".tmx",
```

hooks/conan-center.py
```diff
@@ -634,11 +634,17 @@ def test(out):
...
+                             ".proto", ".capnp", ".c", ".cc", ".c++", ".cpp", ".cxx", ".c++m",
+                             ".cppm", ".cxxm", ".h++", ".hh", ".hxx", ".hpp", ".qrc", ".pro", ".build"]
         recipe_folder = os.path.dirname(conanfile_path)
         for root, _, files in os.walk(recipe_folder):
             if os.path.relpath(root, recipe_folder).replace("\\", "/").startswith("test_package/build"):
                 continue
             for filename in files:
+                if not any(filename.endswith(ext) for ext in ext_to_be_checked):
```
madebr, line 646: I don't know there's a rule disallowing uppercase characters in filenames?
```suggestion
                if not any(filename.lower().endswith(ext) for ext in ext_to_be_checked):
```

hooks/conan-center.py
```diff
@@ -634,11 +634,17 @@ def test(out):
 
     @run_test("KB-H060", output)
     def test(out):
+        ext_to_be_checked = [".cmake", ".conf", ".cfg", ".diff", ".md", ".patch", ".py", ".txt",
+                             ".yml", ".am", ".xml", ".json", ".in", ".ac", ".tsx", ".tmx"
+                             ".proto", ".capnp", ".c", ".cc", ".c++", ".cpp", ".cxx", ".c++m",
+                             ".cppm", ".cxxm", ".h++", ".hh", ".hxx", ".hpp", ".qrc", ".pro", ".build"]
```
madebr, line 640: `.s` and `.asm` might also be useful for assembly files.

ericLemanissier:
> > Wouldn't it be better to do the test as post_export, and check only what is in the local cache ?
> 
> local cache doesn't contain test_package
What is the risk with test_package having CRLF ? It seemed to me that the objective is to avoid at all cost recipe hash modifications caused by line ending change. test_package does not impact recipe hash right?
> 
> > EDIT: allowing .patch and .diff files to have CRLF is a big hole in the net, because a lot of recipe exports at least one of these
> 
> We don't allow, maybe that variable name is bad, I'll update it.
ok, I should have read the code more thoroughly, but I definitely understand it better with the new name, thanks !

hooks/conan-center.py
uilianries: No, good catch!

hooks/conan-center.py

uilianries:
> What is the risk with test_package having CRLF ? It seemed to me that the objective is to avoid at all cost recipe hash modifications caused by line ending change. test_package does not impact recipe hash right?

For now I don't think is a problem, but we are doing it to avoid that problem related to Jenkins. So I think it doesn't harm if we cover one more folder to possible avoid future problems.

madebr (approved):

SpaceIm (approved):

danimtb (approved):

jgsogo (approved):

PR #318, opened 2021-07-02
CRLF - Turn into a warning

jgsogo:
I feel like this is not working as expected: https://github.com/conan-io/conan-center-index/pull/6133

Until we have a better understanding of what is going on and why these PRs are failing, this cannot block PRs.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -648,7 +648,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                     continue
                 lines = open(os.path.join(root, filename), 'rb').readlines()
                 if any(line.endswith(b'\r\n') for line in lines):
-                    out.error("The file '{}' uses CRLF. Please, replace by LF."
+                    out.warn("The file '{}' uses CRLF. Please, replace by LF."
                               .format(filename))
 
 


AndreyMlashkin:
The hook is working as expected. I have just force-pushed, overriding my initial commit.

jgsogo:
We are running some tests on a private repo we have before deploying changes to the library and it is failing for some recipes (`fmt` and `nlopt`). We are afraid that when the repository is cloned in Windows some files will use CRLF locally... or after copying files to the Conan cache in a Windows machine.

Then we saw your PR failing and we thought it could be something _generalized_. Sure we need to wait for your PR.

danimtb (approved):

uilianries:
Have to update the tests. They expect WARN message, not ERROR.

jgsogo:
There are other PRs that are building in windows right now and they are not failing with this error. Not sure if this is something that has just happened to you, @danimtb

jgsogo:
Ok, I'm now understanding all of this better. The hook is checking many more files than the ones we are enforcing in `.gitattributes` so it is possible that there are existing files with CRLF and the hook will complain. This can be inconvenient if the user doesn't know how to change the line endings...

uilianries:
@jgsogo Could we update the hook message with a command hint? Or maybe Use KB error in CCI page

jgsogo:
Hi! In the end, it looks like the error was on one of the Windows machines that was configured with `core.autocrlf=true` (Git configuration), it was converting line-endings and the hook was triggering the error 😌

PR #328, opened 2021-08-05
#312 Validate minimum Cmake version when cxx_std_11 is declared

### LINKED ISSUE #312: Check CMake version when using `cxx_std_##`
Setting are not correctly set when the attribute keys are not supported.

- `cxx_std_20` is `cmake_minimum_required(VERSION 3.12)`
- `cxx_std_17` needs 3.8

_Originally posted by @prince-chrismc in https://github.com/conan-io/conan-center-index/pull/5986#discussion_r654804316_

uilianries:
CMake CXX features for cxx_std are available since CMake 3.8

https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

/cc @prince-chrismc
closes #312

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -59,6 +59,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H058": "ILLEGAL CHARACTERS",
              "KB-H059": "CLASS NAME",
              "KB-H060": "NO CRLF",
+             "KB-H063": "CMAKE REQUIRED VERSION - CXX_STD",
              }
 
 
@@ -666,6 +667,25 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                 if any(line.endswith(b'\r\n') for line in lines):
                     out.error("The file '{}' uses CRLF. Please, replace by LF."
                               .format(filename))
+
+    @run_test("KB-H063", output)
+    def test(out):
+        def validate_cmake_cxx_std(cmakefile_path):
+            cmake_content = tools.load(cmakefile_path)
+            cxx_std = re.search(r"cxx_std_\w+", cmake_content)
+            if cxx_std:
+                match = re.search(r"cmake_minimum_required\(version [\"']?(\w\.\w+)", cmake_content.lower())
+                if match and tools.Version(match.group(1)) < "3.8":
+                    out.error("The CMAKE_CXX_KNOWN_FEATURES ({}) requires CMake 3.8 at least."
+                              " Update the CMakeLists.txt for minimum version required."
+                              .format(cxx_std.group(0)))
+
+        dir_path = os.path.dirname(conanfile_path)
+        test_package_path = os.path.join(dir_path, "test_package")
+        for cmake_path in [os.path.join(dir_path, "CMakeLists.txt"),
+                           os.path.join(test_package_path, "CMakeLists.txt")]:
+            if os.path.exists(cmake_path):
+                validate_cmake_cxx_std(cmake_path)
 
 
 @raise_if_error_output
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -667,6 +668,25 @@ def test(out):
                     out.error("The file '{}' uses CRLF. Please, replace by LF."
                               .format(filename))
 
+    @run_test("KB-H063", output)
+    def test(out):
+        def validate_cmake_cxx_std(cmakefile_path):
+            cmake_content = tools.load(cmakefile_path)
+            cxx_std = re.search(r"cxx_std_\w+", cmake_content)
+            if cxx_std:
+                match = re.search(r"cmake_minimum_required\(version [\"']?(\w\.\w+)", cmake_content.lower())
```
prince-chrismc, line 677: I would have expected digits not words 🤔 

```suggestion
            cxx_std = re.search(r"cxx_std_\d+", cmake_content)
            if cxx_std:
                match = re.search(r"cmake_minimum_required\(version [\"']?(\d\.\d+)", cmake_content.lower())
```

prince-chrismc (commented):
`cxx_std_20` was added in in 3.12 https://cmake.org/cmake/help/v3.12/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

but that's extra

hooks/conan-center.py
uilianries: Word is equivalent to [A-Za-z0-9_], but yeah, digits should fit better.

prince-chrismc (approved):

hooks/conan-center.py
```diff
@@ -59,6 +59,7 @@
              "KB-H058": "ILLEGAL CHARACTERS",
              "KB-H059": "CLASS NAME",
              "KB-H060": "NO CRLF",
+             "KB-H063": "CMAKE REQUIRED VERSION - CXX_STD",
```
Croydon, line 62: Please no more extra hooks for the minimum CMake version required. I think we should combine them all into a single one. Either now or with Conan 2.0 when we increase the CMake version higher than all those hooks require, but probably sooner is better

SSE4:
I am personally tentative if we want to maintain that... perhaps, no?
this would eventually evolve into code validating all the cmake statements, variables, functions, etc. match the minimum required version. I don't think it's our (hook's or CCI's) job to validate that.
the list will keep growing for each new CMake release.
if library requires some CMake release, it just won't build without it - so there is no point to additionally validate it ourselves?
in general, this approach is very questionable for me, I don't see a huge benefit, only snowball of slowly adding more and more CMake variables to check.

jgsogo:
I wonder why CMake doesn't raise an error for this kind of thing. I would like CMake to fail if something is not known to that version... Does anyone know any reason for it? I really don't like adding checks that should be done by someone else.

As a proposal, and (I think) aligned with other's comments: let's keep `conan-center` hook for things that are important/required to ConanCenter... and maybe we can start a hook called `cmake-xxxx` that performs all kinds of checks related to CMake. We can activate both hooks in ConanCenter, or maybe we just activate `cmake-xxx` when running this job https://github.com/conan-io/conan-center-index/issues/2232 to make less noise in pull-requests.

uilianries:
It would be great an API support from Cmake or even a better internal checker. Let's shelve the PR

SSE4 (changes requested):
right now, I don't think it's a job of conan center hook to validate CMake functions used by the library or recipe.

SSE4:
that's an obscure nature of CMake, and it's by design, it will never complain about variables it doesn't know.
if `CMAKE_CXX_KNOWN_FEATURES` is not known to older CMake release, it will be just assumed an empty value, and CMake will run normally to some degree.
that's how they decided about syntax from the beginning, and it's different from C++ any other languages. perhaps, it allows to write code faster without thinking too much about details like initializing everything.
e.g:
```
message(STATUS "${SOME_VARIABLES_THAT_DOES_NOT_EXIST}")
```
will never abort the CMake execution, and will never trigger a single warning. it will just assume `SOME_VARIABLES_THAT_DOES_NOT_EXIST` is an empty string. that's why it cannot be validated easily, I believe.
so if future CMake version adds a variable named `CMAKE_CXX_KNOWN_FEATURES`, and your CMakeLists.txt uses it, and you run old CMake release - nothing happens.

PR #335, opened 2021-08-12
[conan-center] Validate topics attribute

### LINKED ISSUE #334: [conan-center] Invalid topics should be validated
Most CCI recipes are using "conan" as topic, but the real usage for topics is related to tags, for filtering and search. Thus, "conan" is a generic and invalid topic.

Recipe should the upstream's tags from Github when available, or use related keywords present in the project description.

uilianries:
Due the high number of recipes using "conan" as topic, we can not set this hook as error, it would break many packages in Conan Center. Let's use Warning level, which is good enough.

close #334

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -59,7 +59,8 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H058": "ILLEGAL CHARACTERS",
              "KB-H059": "CLASS NAME",
              "KB-H060": "NO CRLF",
-             "KB-H062": "TOOLS CROSS BUILDING"
+             "KB-H062": "TOOLS CROSS BUILDING",
+             "KB-H064": "INVALID TOPICS",
              }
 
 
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -682,6 +683,16 @@ def _check_content(content, path):
             test_package_content = tools.load(test_package_path)
             _check_content(test_package_content, test_package_path)
 
+    @run_test("KB-H064", output)
+    def test(out):
+        topics = getattr(conanfile, "topics")
+        if topics and isinstance(topics, (list, tuple)):
+            invalid_topics = ["conan"]
```
SSE4, line 690: can you explain why is it invalid?
the topic is totally valid and indicates the recipe/package/repository is compatible with conan.
if someone is populating github (or similar) tags from conan topics with scripts, it's useful.

SSE4: for me personally, that's super useful as I can search on github for topic conan (plus some search string, like "multimedia") and find repositories with conan recipes.
I guess same is valid for other systems (bitbucket, gitlab, etc.)
if someone is forking CCI recipe to his own repo, he gets topics as well, so I don't see why it should be removed.

Croydon: > if someone is populating github (or similar) tags from conan topics with scripts, it's useful.

That was the initial idea why at Bincrafters we starting adding the tag `conan`, but it seems redundant. If you read topics from a Conan recipe you already know that it is a Conan recipe. So scripts that can populate e.g. GitHub tags can add the `conan` tag always.

SSE4 (approved):

madebr (approved):

danimtb (approved):

PR #336, opened 2021-08-14
Allow 'share' folder in package IF it is added to self.cpp_info.resdirs

madebr:
For https://github.com/conan-io/conan-center-index/pull/6852

Also, https://github.com/conan-io/conan-center-index/pull/6846#issuecomment-898749915


uilianries:
I prefer renaming and patching instead of adding this flexibility. As we have a pattern, I know if I check res/ folder, I'll find resource files. Besides that, that PR is a rare case.

sparik:
@uilianries KB-H013 has a whitelist of recipes. We can also add `gettext` to the whitelist

PR #342, opened 2021-09-01
don't allow CMAKE_CXX_STANDARD

ericLemanissier:
Forcing CMAKE_CXX_STANDARD in a cmakelists overrides conan setting compiler.cppstd, so it creates an incoherent graph when consuming recipe. there are currently 64 offending recipes: https://github.com/conan-io/conan-center-index/search?l=CMake&q=CMAKE_CXX_STANDARD

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -61,6 +61,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H060": "NO CRLF",
              "KB-H062": "TOOLS CROSS BUILDING",
              "KB-H064": "INVALID TOPICS",
+             "KB-H065": "CMAKE CXX STANDARD"
              }
 
 
@@ -692,6 +693,20 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                 if topic in invalid_topics:
                     out.warn("The topic '{}' is invalid and should be removed from topics "
                              "attribute.".format(topic))
+
+    @run_test("KB-H065", output)
+    def test(out):
+        def check_for_cxx_standard_flag(cmakelists_path):
+            cmake_content = tools.load(cmakelists_path)
+            if "CMAKE_CXX_STANDARD" in cmake_content.lower():
+                out.error("The CMake definition 'set(CMAKE_CXX_STANDARD ...)' is not allowed. "
+                          "Remove it from {}.".format(os.path.relpath(cmakelists_path)))
+
+        dir_path = os.path.dirname(conanfile_path)
+        for cmake_path in [os.path.join(dir_path, "CMakeLists.txt"),
+                           os.path.join(dir_path, "test_package", "CMakeLists.txt")]:
+            if os.path.exists(cmake_path):
+                check_for_cxx_standard_flag(cmake_path)
 
 
 @raise_if_error_output
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -700,7 +700,9 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
             cmake_content = tools.load(cmakelists_path)
             if "CMAKE_CXX_STANDARD" in cmake_content.lower():
                 out.error("The CMake definition 'set(CMAKE_CXX_STANDARD ...)' is not allowed. "
-                          "Remove it from {}.".format(os.path.relpath(cmakelists_path)))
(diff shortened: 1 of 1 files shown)


uilianries:
And what's the alternative?

ericLemanissier:
use validate() to raise if the standard is too low

uilianries:
please, add it to the hook message

ericLemanissier:
I'm ready to do all the pull requests to fix recipes before this one is merged, if we agree on the principle

SpaceIm:
> And what's the alternative?

Patch upstream CMakeLists to add `target_compile_features()` with minimum required standard, instead of adding CXX_STANDARD in CMakeLists wrapper.

ericLemanissier:
AFAIU, `target_compile_features` is not a solution, because cmake automatically adds the needed compilation flags, which mean you still end up having some parts of your tree built with a certain standard, and other parts with another standard

SpaceIm:
> AFAIU, `target_compile_features` is not a solution, because cmake automatically adds the needed compilation flags, which mean you still end up having some parts of your tree built with a certain standard, and other parts with another standard

CMake uses both CXX_STANDARD and cxx features to know which C++ standard to inject.
When compiler.cppstd is not set in conan profile, conan doesn't inject CXX_STANDARD, so consumer likely want to be driven by the behavior of cxx feature which is the best practice.
When compiler.cppstd is set in conan profile, each recipe will evaluates if this standard is sufficient in validate(), raise eventualy, and if it passes the test, CXX_STANDARD will take precedence over cxx features, so it's fine.

ericLemanissier:
I *think* that more than 1% of the packages are affected by this problem, but I don't have yet precise figures. Is there somewhere a list of the most downloaded packages on CCI, so that I can go check the sources of the matching project ? Until now we now that at least gtest, fmt and asio have this problem of api/abi changing with cppstd.

madebr:
systemc is also an offender.
It embeds `__cplusplus` in the namespace.

ericLemanissier:
> * Because it doesn't bring any improvement.

sorry for repeating, but there are at least two improvements:

1. Apple-clang (which defaults to c++98) users would get a clear error message instead of a cryptic error message, as soon as they consume a package requiring c++11 or later. (same for other compilers, but with an even more recent standard)
2. systemc/fmt/asio/gtest/... consumers would not risk a cryptic ABI/API break if any of there dependencies forces c++XX and some other forces C++YY

SpaceIm:
> > * Because it doesn't bring any improvement.
> 
> sorry for repeating, but there are at least two improvements:
> 
>     1. Apple-clang (which defaults to c++98) users would get a clear error message instead of a cryptic error message, as soon as they consume a package requiring c++11 or later. (same for other compilers, but with an even more recent standard)

I'm using quite often apple-clang without compiler.cppstd set, and never had cryptic error. It just works for the vast majority of recipes, and I can use pre-build packages. Here you'll prevent to build all packages requiring C++11 or higher on Macos in CI of CCI.

ericLemanissier:
For cryptic error message example, please see https://github.com/conan-io/conan-center-index/pull/7236#issuecomment-916878176

jgsogo:
This is IMHO the biggest problem we have now in ConanCenter. But the issue comes not only from our `CMakeLists.txt` wrapper, but also from the library itself. I think there are lots of libraries that will ignore a value provided from outside... the last one I came across was `google-cloud-cpp` and just because I'm using it for a C++ project where I really needed to force a `compiler.cppstd` value.

When I think about this issue I usually come back to this: https://github.com/conan-io/examples/pull/73. If I think in terms of CCI, it would _easy_ to fix it if we run the builds for all the different `compiler.cppstd` values (stable ones), but we cannot afford it in terms of time/resources.
Next thought, how to reduce the number of configurations? Using compatible_packages, no need to build all of them, only the ones that generate different ABIs... but that requires deep internal knowledge of the library (not easy). And it also requires a way to filter packageIDs that we don't need to generate because they will be covered by a compatible_package.

I'll follow this thread with interest.

Minimonium:
The main issue the last time we discussed it was that you can't simply rely on compatible_packages (the POC of a function that extends `check_min_cppstd` I wrote a year ago does use it) - it's that you need to propagate that compatibility to consumers. 

If someone uses Boost (in their public interfaces) - it's no longer possible for the to mix C++14 and C++17 in the same TU. A solution requires native Conan support.

PR #347, opened 2021-09-16
Exclude mono from default package layout.

Pro:
See https://github.com/conan-io/conan-center-index/pull/7239#issuecomment-920732232

For mono we need to keep the `etc` directory.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -903,7 +903,7 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
 
     @run_test("KB-H013", output)
     def test(out):
-        if conanfile.name in ["cmake", "android-ndk", "zulu-openjdk", "mingw-w64", "openjdk"]:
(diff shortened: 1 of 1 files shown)


Croydon (approved):

uilianries:
Do you really need `etc/` folder, or can it be moved to `res/` folder? By default Conan uses `res`, which is already listed by Conan generators.

Pro:
@uilianries I tested this by just moving the files to the res folder, but then I get the same error as mentioned in the linked comment.

It seems that mono has somewhere hardcoded to look for files in etc folder.

Just using Github search, there are quite a few places:
https://github.com/mono/mono/search?q=etc

E.g., this one:
https://github.com/mono/mono/blob/bdd772531d379b4e78593587d15113c37edd4a64/runtime/Makefile.am

And I would like to avoid that we need to patch all these places, since maintining these patches will be a lot of effort.

uilianries:
@Pro Thank you for checking. Indeed we need a exception rule for this package.

uilianries (approved):

danimtb (approved):

PR #348, opened 2021-09-30
skip emsdk from kb-h016

uilianries:
Related to https://github.com/conan-io/conan-center-index/pull/6163

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -965,7 +965,7 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
     @run_test("KB-H016", output)
     def test(out):
         if conanfile.name in ["cmake", "msys2", "strawberryperl", "pybind11", "ignition-cmake",
-                              "extra-cmake-modules"]:
+                              "extra-cmake-modules", "emsdk"]:
             return
         bad_files = _get_files_following_patterns(conanfile.package_folder, ["Find*.cmake",
                                                                              "*Config.cmake",


jgsogo (dismissed):

SSE4 (dismissed):

SSE4:
@uilianries CI is not happy :(

jgsogo:
Why this is not taking into account the changes from `master`? It should be using the right `pylint` version.

uilianries:
Good morning fellas, yes, it seems related to pylint version.

jgsogo:
but this one (already merged) should have been fixed it: https://github.com/conan-io/hooks/pull/349. The problem is that this PR is not using the requirements that are listed in the file 🤷

uilianries:
@jgsogo we should use `--no-cache-dir`, otherwise `pip` will be prefer its cache, just like Conan.

.ci/generate_env_linux.sh
```diff
@@ -39,5 +39,5 @@ esac
 pyenv activate conan
 python --version
 pip install --upgrade pip
-pip3 install --requirement .ci/requirements_linux.txt
+pip3 install --no-cache-dir --requirement .ci/requirements_linux.txt
```
jgsogo, line 42: This should NOT be needed, unless `pip3` is caching file `.ci/requirements_linux.txt` which would be something absolutely unexpected IMHO.

uilianries: i gonna revert

jgsogo (approved):
Finally, it was just a lazy CI 😄

uilianries:
It's Friday, the CI wants a break!

PR #350, opened 2021-10-01
[tests] Move py versions

jgsogo:
Let's keep as a general rule the following python versions:
 * (not strictly required) `py27` until Conan stops testing it or we can no longer run this tests
 * `py36` minimum version required according to tribe. **It is also the version we run in C3i**
 * `py<latest>`. Latest python version, it is likely the one used by most of consumers. 

---

These tests are running in the Linux image of Conan CI. That container hasn't been regenerated for a while and available Python versions are quite old (cc/ @czoido)... I'm using the same ones I see here https://github.com/conan-io/conan_ci_jenkins/blob/master/resources/org/jfrog/conanci/python_runner/conf.py#L19 because they are already available in the image and we don't need to install/build them each time. Alternatively, if we really want to run the latest patch version, we should use our own image and keep it updated.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -35,30 +35,61 @@ pipeline {
                         }
                     }
                 }
-                stage('Windows - py38') {
+                stage('Windows - py36') {
                     agent {
                         label 'Windows'
                     }
                     stages {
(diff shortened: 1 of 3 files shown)


uilianries (approved):
Agreed

PR #351, opened 2021-10-01
Skip emsdk from KB-H020

jgsogo:
For https://github.com/conan-io/conan-center-index/pull/6163

Two things I'm learning here:

- [ ] Probably we should maintain a list of _cross-compilers_ so they are all added to a list, and the same list is used in all the hooks that apply to them.
- [ ] (To check) All the hooks run and print their message before raising the error?

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -953,7 +953,7 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
 
     @run_test("KB-H020", output)
     def test(out):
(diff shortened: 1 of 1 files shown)


Croydon:
Both things would be certainly improvements 👍 

The last one would even save some CI time and iterations for CCI, where often people encounter one error, then the next, then the next... from the hooks

Croydon (approved):

uilianries (approved):
Déjà vu

Issue #354, opened 2021-10-03
[conan-centre] forbid CONAN_LIBS and CONAN_PKG


SSE4:
in the near future (conan 2.0), old CMake generators (`cmake`, `cmake_multi`, `cmake_paths`, etc.) will be removed, as well as `conan_basic_setup`.
targets `CONAN_PKG::` won't be provided any longer.
also, variables like `CONAN_LIBS` won't be provided as well.
these things were mainly used by `test_package`, and sometimes in patches to provide conan integration.
it's time to start removing them.

PR #355, opened 2021-10-03
- forbid CONAN_PKG and CONAN_LIBS

SSE4:
closes: #354 

we will need to run hooks against current conan-center to understand the scale of the destruction

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -64,6 +64,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H061": "NO BUILD SYSTEM FUNCTIONS",
              "KB-H062": "TOOLS CROSS BUILDING",
              "KB-H064": "INVALID TOPICS",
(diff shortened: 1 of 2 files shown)


Croydon (commented):
I have concerns with this in cases where upstream does not provide official targets. This will force CCI to invent targets and make downstream depend on them. Having it somehow namespaced to Conan makes it clearer

jgsogo:
@Croydon, this is aligned with Conan milestones and it is strategic. You can find a longer rationale here: https://github.com/conan-io/conan-center-index/pull/7153 

TL;DR.- Conan v2.0 will rely on `CMakeDeps` generator, and only regular targets will be available (and documented). We need to approach Conan v2.0 as much as possible so we can make recipes compatible with both Conan versions.

Of course, like everything, if you have good reasons to propose something different, it is time to suggest and provide feedback in the conan repository or the tribe. 

Thanks for your understanding.

jgsogo:
(Executing a dry run for this branch. I'll post output as soon as it is available)

Issue #308, opened 2021-06-10
yaml linter


ericLemanissier:
We should consider linting conandata.yml and config.yml files, for example using [yamllint](https://yamllint.readthedocs.io/en/stable/index.html)
Example of case where it would have save some non negligible debugging time: https://github.com/conan-io/conan-center-index/commit/61eb40724ee06218ebad43f2e20fd51455281c51#diff-c72439ef9b3281fe119686121ba38bb839dc40459798c2641142eb4135f15f61R16

exemple of execution with yamllint:
test.yml:
```
patches:
  a
  b
patches:
  c
```
```
$ yamllint --no-warnings test.yml
test.yml
  4:1       error    duplication of key "patches" in mapping  (key-duplicates)
```

uilianries:
> I like the idea, but the problem with this is that the CCI hooks don't require any other Python package so far and requiring dependencies makes handling them more complex

True.

Also, we avoided linter because not all contributors know Python, I believe most people are C/C++ devs. So if we start to block because of linters, probably more users will give up first PR.

I would consider it as a recommendation in Docs, and of course, we can implement yaml linter hook, but not mandatory for CCI. I totally understand the trade off situation, we could avoid more errors in CCI.

ericLemanissier:
> I like the idea, but the problem with this is that the CCI hooks don't require any other Python package so far and requiring dependencies makes handling them more complex

Yes, this bothers me too. I searched for another linter directly callable from python code, but could not find one. I still posted the issue because maybe *someone* here knows one ?

> Also, we avoided linter because not all contributors know Python, I believe most people are C/C++ devs. So if we start to block because of linters, probably more users will give up first PR.
> 
> I would consider it as a recommendation in Docs, and of course, we can implement yaml linter hook, but not mandatory for CCI. I totally understand the trade off situation, we could avoid more errors in CCI.

Just to be clear, the purpose is not to catch style errors, nor smelly code, but real errors with zero false positives.
e.g.: if someone adds a "patch" key to a config.yml which already has one, the net result is that all the patches mentioned in the original key are silently ignored. (which is what happened in https://github.com/conan-io/conan-center-index/commit/61eb40724ee06218ebad43f2e20fd51455281c51#diff-c72439ef9b3281fe119686121ba38bb839dc40459798c2641142eb4135f15f61R16)

uilianries:
[...]

PR #358, opened 2021-10-06
imlement yaml linter hook

ericLemanissier:
fixes https://github.com/conan-io/hooks/issues/308

### DIFF
--- a/hooks/yaml_linter.py
+++ b/hooks/yaml_linter.py
@@ -1,0 +1,41 @@
+# coding=utf-8
+
+import os
+import platform
+import subprocess
+
+from conans.errors import ConanException
+from conans.tools import logger
+
+
+CONAN_HOOK_YAMLLINT_WERR = "CONAN_YAMLLINT_WERR"
+
+
+def pre_export(output, conanfile_path, *args, **kwargs):
+    output.info("Lint yaml '{}'".format(conanfile_path))
+    conanfile_dirname = os.path.dirname(conanfile_path)
+    conandata_file = os.path.join(conanfile_dirname, 'conadata.yaml')
+
+    lint_args = ['-f', 'parsable',
+                 '-d', '"{extends: default, rules: {document-start: disable}}"']
+
+    try:
+        command = ['yamllint'] + lint_args + [conandata_file.replace('\\', '/')]
+        command = " ".join(command)
+        shell = bool(platform.system() != "Windows")
+        p = subprocess.Popen(command, shell=shell, bufsize=10,
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        yamllint_stdout, yamllint_stderr = p.communicate()
+        yamllint_stdout = yamllint_stdout.decode('utf-8')
+    except Exception as exc:
+        output.error("Unexpected error running linter: {}".format(exc))
+        return
+    errors = 0
+    for line in messages.splitlines():
+        output.info(line)
+        parts = line.split(' ')
+        errors += int(parts[1] == "[error]")
+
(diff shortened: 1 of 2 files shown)


ericLemanissier:
> the linter can be very useful. The main downside is that we don't have a way to install external dependencies for hooks...

I just added some code that made it explicit what has to be done if yamllint is not installed.

> Another question: Should this be part of the conan-center hook or better keep it as a different hook for wider use rather than ConanCenter only?

As it has an external requirement, I think we cannot include it in conan-center hooks. That said we can still run the hook in Conan-center-index CI, as demonstrated in https://github.com/ericLemanissier/conan-center-index/pull/5


> Anyway, if this is going to check the conandata.yml only I'd probably change the name to explicit this or make the hook more generic to check all YAML files exported alongside the recipe.

ok, I guess that would make the check a post_export instead of pre_export

ericLemanissier:
How do we know If it is cci's structure ? We could end up scanning the full disk

hooks/yaml_linter.py
```diff
@@ -0,0 +1,66 @@
...
+        "empty-lines": "{level: warning}",
+        "indentation": "{level: warning}",
+        "trailing-spaces": "{level: warning}",
+    }
+
+    lint_args = ['-f', 'parsable',
+                 '-d', '"{extends: default, rules: {%s}}"' %
+                 ", ".join("%s: %s" % (r, rules[r]) for r in rules)]
+    lint_args.append('"%s"' % conanfile_dirname.replace('\\', '/'))
+    configfile = os.path.join(conanfile_dirname), "..", "config.yml")
```
uilianries, line 39:
```suggestion
    configfile = os.path.join(conanfile_dirname, "..", "config.yml")

```

Ops!

uilianries:
Well, it will be funny once merged. Not in Conan Center by default, but still has a big potential:

```
[HOOK - yaml_linter.py] pre_export(): Lint yaml '/home/uilian/Development/conan/conan-center-index/recipes/zlib/1.2.11/conanfile.py'
[HOOK - yaml_linter.py] pre_export(): /home/uilian/Development/conan/conan-center-index/recipes/zlib/1.2.11/conandata.yml:4:10: [warning] wrong indentation: expected 6 but found 9 (indentation)
[HOOK - yaml_linter.py] pre_export(): /home/uilian/Development/conan/conan-center-index/recipes/zlib/1.2.11/conandata.yml:5:10: [warning] wrong indentation: expected 6 but found 9 (indentation)
[HOOK - yaml_linter.py] pre_export(): /home/uilian/Development/conan/conan-center-index/recipes/zlib/1.2.11/conandata.yml:6:8: [warning] wrong indentation: expected 4 but found 7 (indentation)
[HOOK - yaml_linter.py] pre_export(): YAML Linter detected '0' errors
```

tox.ini
```diff
@@ -33,3 +33,5 @@ commands =
     coverage: coverage combine
     coverage: coverage report
     coverage: coverage xml
+    
+recreate = true
```
uilianries, line 37: Why do you need recreate? I think it should be optional, you still can pass to tox when running `tox --recreate`

ericLemanissier: It's not mandatory in the long term indeed, but this change https://github.com/conan-io/hooks/pull/358/files#diff-080a90ef1f86f679fc01d7d923bafc8a3472927c4727cd8fd961e5af1717e79aR9 generates an environment different from the one needed on master currently, so the recreation has to be forced at some point, in order to make this PR pass.

jgsogo (approved):
Feel free to add an entry to README.md

PR #359, opened 2021-10-07
Move KB-H047 - Non ASCII

uilianries:
As we are running only Python 3 and C3I requires Python <= 3.6, we can accept non-ascii characters in recipes.

I also removed Python 2.7 testing for Windows. Why? We no longer run Python 2 on CCI.

UPDATE:

Now non-ascii is a separated hook.


SSE4:
you can remove py27 altogether from CI (e.g. tox.ini and whenever else it's mentioned)

SSE4 (approved):

SSE4 (approved):

PR #363, opened 2021-10-13
Skip OpenTelemetry Proto from hook 14

uilianries:
Required by https://github.com/conan-io/conan-center-index/pull/7678 

That package does not contain source files, only .proto files.


uilianries:
Needed again

jgsogo (approved):

PR #377, opened 2022-02-01
- [conan-center] skip hook messages for dependencies

SSE4:
it's pretty annoying that conan-center hook reports not only errors/warnings for the package being built/tested, but for all of its dependencies. they are obviously useless and uninteresting for the contributor.

it's mainly for the `post_package_info` hook right now, as other methods besides `package_info` (`export`, `source`, `build`, `package`) are only executed for the package being built/tested.

from the signature of the hook, there is no easy way to say if current conanfile/reference is the main one, or dependency:
```
def post_package_info(output, conanfile, reference, **kwargs):
```
there is nothing interesting in public `ConanFile` fields as well.

to complicate things, it seems like the hooks object is being re-created each time, so it's not trivial to save any state between method invocations, it's always brand new, constructed from scratch.

the reliable process-wide storage is `sys.modules`, which works well, regardless of how many times python file is loaded.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -864,6 +864,7 @@ def pre_build(output, conanfile, **kwargs):
 
 @raise_if_error_output
 def post_package(output, conanfile, conanfile_path, **kwargs):
+    this.reference = str(conanfile)
     @run_test("KB-H012", output)
     def test(out):
         if conanfile.version == "system":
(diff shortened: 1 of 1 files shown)


uilianries (dismissed):

SSE4:
it's a bit complicated, as we don't invoke `conan create` in CI, we do `conan install --build=<package>` and then `conan test`. so `pre_export` is not called in the same process as `post_package_info`. I'll change it.

uilianries (approved):
LGTM

Issue #370, opened 2021-11-23
[conan-centre] fix hooks for develop2 (2.0 alpha)


SSE4:
we will (soon) need to fix hooks for develop2 (2.0 alpha), right now they always fail with some import errors.

PR #380, opened 2022-02-02
- [conan-center] compatibility with develop2 (alpha)

SSE4:
closes: #370


uilianries:
We could run Conan develop2, but skipping errors, just in case. But it's for another PR.

uilianries (approved):

SSE4:
> We could run Conan develop2, but skipping errors, just in case. But it's for another PR.

yes, I tried to run tests, but they have more errors. it would be nice if we can run CI against develop2 alpha as well.

PR #382, opened 2022-02-07
Restrict tools.os_info usage

### LINKED ISSUE #373: [conan-center] KB-H061 didn't detect os_info in config_options
from https://github.com/conan-io/conan-center-index/pull/7535
the code is:
```
    def config_options(self):
        if not os_info.is_windows:
            self.options.enable_wininet_for_http = False

        if os_info.is_windows:
            del self.options.fPIC
```
which results is an Unexpected Error due to the different package ID computed on Windows and Linux nodes.

uilianries:
The `tools.os_info` is used only for specific cases on Conan Center Index:
- Packages related to Linux system packages
- `AutoToolsBuildEnvironment(self, win_bash=tools.os_info.is_windows)`

https://docs.conan.io/en/latest/reference/conanfile/methods.html#system-requirements

This hook restricts its usage with `config_options` and `configure`. Maybe we could consider something more aggressive, like, only accepting `os_info` on `build` method and protected methods.

closes #373

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -69,6 +69,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H062": "TOOLS CROSS BUILDING",
              "KB-H064": "INVALID TOPICS",
              "KB-H065": "NO REQUIRED_CONAN_VERSION",
+             "KB-H066": "NO OS_INFO",
              }
 
 
@@ -760,6 +761,14 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
             out.warn("tools.get with strip_root=True is available since Conan {0}. "
                      "Please add `required_conan_version >= \"{0}\"`"
                      "".format(str(required_version)))
+
+    @run_test("KB-H066", output)
+    def test(out):
+        for method in ["config_options", "configure"]:
+            method_obj = getattr(conanfile, method)
+            if "os_info." in inspect.getsource(method_obj):
+                out.error("tools.os_info must not be used in 'config_options' and 'configure'."
+                          " Replace by 'self.settings.os'.")
 
 
 @raise_if_error_output
--- a/tests/test_hooks/conan-center/test_tools_os_info.py
+++ b/tests/test_hooks/conan-center/test_tools_os_info.py
@@ -1,0 +1,51 @@
+import os
+import textwrap
+
+from conans import tools
+
+from tests.utils.test_cases.conan_client import ConanClientTestCase
+
+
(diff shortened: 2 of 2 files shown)


ericLemanissier:
cannot this be merged into existing https://github.com/conan-io/hooks/blob/fec9d30aa20e08d7fd3be1cb75eb56f4001b5327/hooks/conan-center.py#L658 ?

uilianries:
> cannot this be merged into existing
> 
> https://github.com/conan-io/hooks/blob/fec9d30aa20e08d7fd3be1cb75eb56f4001b5327/hooks/conan-center.py#L658
> ?
> 
> EDIT: AFAIU, it should already flagged:
> 
> https://github.com/conan-io/hooks/blob/fec9d30aa20e08d7fd3be1cb75eb56f4001b5327/hooks/conan-center.py#L665-L666

You and SSE4 are right, indeed Hook 61 should catch it. But Conan Center hook is now >1k2 lines, I can remember everything :sweat_smile: . Let me check it. Thank you for spotting it!

hooks/conan-center.py
```diff
@@ -685,7 +685,8 @@ def visit_Attribute(self, node):
                 methods_stack_no_build_info_allowed = [fdef for fdef in self.function_def_stack if fdef in self.METHODS_NO_BUILDINFO]
                 if methods_stack_no_build_info_allowed:
                     # FIXME: Not all python 2.7 interpretors have node.end_lineno or node.end_col_offset
-                    if node.attr == "os_info" and isinstance(node.value, ast.Name) and node.value.id == "tools":
+                    if (node.attr == "os_info" and isinstance(node.value, ast.Name) and node.value.id == "tools") or \
+                       ("is" in node.attr and isinstance(node.value, ast.Name) and node.value.id == "os_info"):
```
jgsogo, line 689: this `"is" in node.attr` is matching the `is` in `os_info.is_macos`, `os_info.is_xxx`?

uilianries: Yes, maybe rename it to "is_"

SSE4: maybe it should be `startswith("is_")` then

jgsogo: Can we just test if we are importing `os_info` from `conans.tools`?

SSE4: nope, importing is legit. only used in certain methods is illegal (e.g. config_options), while in other methods it's fine (e.g. build)

jgsogo (approved):
Looks good. I think we can merge it without checking how many recipes might be affected before... there shouldn't be too many 😅

SSE4 (approved):

PR #384, opened 2022-02-10
Alert about windows short paths

### LINKED ISSUE #362: [conan-center] Suggest when short_paths is required
The max path length on Windows is well known, besides the workaround, we still use [short_paths](https://docs.conan.io/en/latest/reference/conanfile/attributes.html#short-paths) attribute to avoid such cases where a package name is too long and the package folder can pass Windows max path.

Some suggestions:

- When the package name is long (more than 12 characters), short_paths should be considered due the user name over C:/Users
- If include folder are part of this limitation, check after package step, if the include folder path exceeded max path length

uilianries:
- We have to consider around 160 characters for Conan folder structure, e.g `C:\Users\foo\.conan\data\nlohmann_json\3.9.1\authorname\testing\package\5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9\include`

closes #362

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1024,9 +1024,9 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
     def test(out):
         conanfile_content = tools.load(conanfile_path)
         if not re.search(r"(\s{4}|\t)short_paths\s*=", conanfile_content):
-            # INFO: Need to reserve around 120 characters for package folder path
+            # INFO: Need to reserve around 160 characters for package folder path
             short_paths_name_length = 12
-            include_max_length_path = 120
+            include_max_length_path = 96
             if len(conanfile.name) >= short_paths_name_length:
                 out.warn(f"The package name '{conanfile.name}' is too long and may exceed Windows max path length. "
                           "Add 'short_paths = True' in your recipe.")
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1026,7 +1026,7 @@ def post_package(output, conanfile, conanfile_path, **kwargs):
         if not re.search(r"(\s{4}|\t)short_paths\s*=", conanfile_content):
             windows_max_path = 256
             # INFO: Need to reserve around 160 characters for package folder path
-            for folder, length in [(conanfile.source_folder, 100), (conanfile.package_folder, 160)]:
+            for folder, length in [(conanfile.source_folder, 140), (conanfile.package_folder, 160)]:
                 file_max_length_path = windows_max_path - length
                 with tools.chdir(folder):
                     for (root, _, filenames) in os.walk("."):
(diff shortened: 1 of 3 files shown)


hooks/conan-center.py
```diff
@@ -1019,6 +1020,23 @@ def test(out):
...
 
+    @run_test("KB-H066", output)
+    def test(out):
+        conanfile_content = tools.load(conanfile_path)
+        if not re.search(r"(\s{4}|\t)short_paths\s*=", conanfile_content):
+            # INFO: Need to reserve around 160 characters for package folder path
+            short_paths_name_length = 12
+            include_max_length_path = 96
+            if len(conanfile.name) >= short_paths_name_length:
+                out.warn(f"The package name '{conanfile.name}' is too long and may exceed Windows max path length. "
```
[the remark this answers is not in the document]
uilianries: Indeed, it could help in case new versions come with longer paths. We have packages with long names, but not many, I checked it today. Anyway, I'll remove this rule and keep only file path.

uilianries:
@ericLemanissier please, review it again.

ericLemanissier (approved):
should we consider a similar pre_build hook, to check the content of the source folder?

uilianries:
> should we consider a similar pre_build hook, to check the content of the source folder?

Good idea. I'll update this PR, we can do all at same hook.

uilianries:
@ericLemanissier finally added source folder rule. The CI cheated this hook, it uses a long path for testing.

hooks/conan-center.py
```diff
@@ -1019,6 +1020,23 @@ def test(out):
             out.error("The conan-center repository doesn't allow Microsoft Visual Studio runtime files.")
             out.error("Found files: {}".format("; ".join(bad_files)))
 
+    @run_test("KB-H066", output)
+    def test(out):
+        conanfile_content = tools.load(conanfile_path)
+        if not re.search(r"(\s{4}|\t)short_paths\s*=", conanfile_content):
+            windows_max_path = 256
+            # INFO: Need to reserve around 160 characters for package folder path
+            for folder, length in [(conanfile.source_folder, 140), (conanfile.package_folder, 160)]:
```
ericLemanissier, line 1029: this is a post_package hook. shouldn't the source be checked in a post_source hook ?

uilianries: Unfortunately we can't duplicate the hook number. Otherwise, we would need to create 2 hooks separated.

SSE4: I don't see any problem if we have multiple hook numbers and several KB entires, like:
KBH066 - short paths (source)
KBH067 - short paths (build)
KBH068 - short paths (package)
(I highly doubt we will ever need export and other folders in practice)

uilianries: It seems it works actually! It tested again and I see no errors now, probably some prone error during development. 
Please, @ericLemanissier and @SSE4 take a look now.

ericLemanissier (approved):

SSE4 (approved):

PR #368, opened 2021-11-04
Test for required_conan_version if strip_root is used KB-H065

theirix:
Recipes are upgraded to use `tools.get` with a [`strip_root=True`](https://docs.conan.io/en/latest/changelog.html#jan-2021). A directive `required_conan_version` can be missing. This hook checks it.

Proposed id: #KB-H065: NO REQUIRED_CONAN_VERSION

PR #395, opened 2022-03-03
Fix false positives for required_conan_version, KB-H065

theirix:
It is a fixup to PR #368 , can block hook usage. The test for passing hook check was incorrectly written so it cannot detect an error in regex. Hook raises false positives to completely valid `required_conan_version = ">=1.33.0"`.


uilianries (approved):

uilianries:
Thank you!

PR #396, opened 2022-03-03
[conan-center] KB-H069 default_options is not allow on test_package

### LINKED ISSUE #287: Avoid usage of 'default_options' in test-package
Seen here: https://github.com/conan-io/conan-center-index/pull/5024

If using `default_options` in `test_package/conanfile.py` it is very likely that the computed package-id won't exist (we only build packages with the default value for options). It would be nice to prevent these tests to modify the default options. Not sure if there is a hook right now that can do the work, maybe we need to introduce it in Conan first.

uilianries:
Test package should not use default options, otherwise, it won't test the default package configuration.

closes #287

Wiki: 

#### **<a name="KB-H069">#KB-H069</a>: "TEST PACKAGE - NO DEFAULT OPTIONS"**

When testing the built package, we need to validate the default configuration, including its options. in case of use
custom options via `default_options`, this will not guarantee the default package state.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -73,6 +73,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H065": "NO REQUIRED_CONAN_VERSION",
              "KB-H066": "SHORT_PATHS USAGE",
              "KB-H068": "TEST_TYPE MANAGEMENT",
+             "KB-H069": "TEST PACKAGE - NO DEFAULT OPTIONS",
              }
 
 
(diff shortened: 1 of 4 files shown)


SSE4 (dismissed):
I think it's reasonable. are there valid use-cases? maybe changing some options for other requirements/build_requirements (not a package under the test)

uilianries:
> I think it's reasonable. are there valid use-cases? maybe changing some options for other requirements/build_requirements (not a package under the test)

Maybe shared option for requires/build_requires only. Otherwise, we would not have that configuration available on Conan Center.
But still, if you need to enforce shared, you should do it on the recipe, not on the test package

uilianries:
Done

SSE4:
> Details

now it fails :(

SSE4:
also, job wasn't able to report into this issue, but here is the output (part of it):
```
FAIL: conan export recipes/backward-cpp/all/conanfile.py backward-cpp/1.6@
FAIL: conan export recipes/backward-cpp/all/conanfile.py backward-cpp/1.5@
FAIL: conan export recipes/backward-cpp/all/conanfile.py backward-cpp/1.4@
FAIL: conan export recipes/gettext/all/conanfile.py gettext/0.21@
FAIL: conan export recipes/gettext/all/conanfile.py gettext/0.20.1@
```

uilianries:
I'll investigate. Thank you @SSE4 for pinging!

uilianries:
@SSE4 The new conan pylint does not like `conans` as import. I just updated to `conan`. Unfortunately, `python_requires`is not ported to `conan` yet.

SSE4:
> @SSE4 The new conan pylint does not like `conans` as import. I just updated to `conan`. Unfortunately, `python_requires`is not ported to `conan` yet.

what about `python_requires`? where is it used and for what and why?

uilianries:
> what about python_requires? where is it used and for what and why?

For testing in that case. As it requires the pip package python_version, it validates a possible import error.

SSE4 (approved):

PR #403, opened 2022-03-08
[conan-center] KB-H071 Validate relocable shared libraries

### LINKED ISSUE #376: [conan-center] Add a hook to check whether installed shared libs are relocatable on Linux & macOS
Currently, many recipes don't produce state of the art relocatable shared libs on macOS.

It would be nice to check whether shared libs on macOS properly have `@rpath/<shared>` in install tree (`@rpath` token is supported since macOS 10.5 Leopard, so since 2007...).
It must not be `<shared>` or `<absolute/path/to/shared>` (many binaries produced by conan-center recipes on macOS have the former actually, it's bad).
see https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html#prop_tgt:MACOSX_RPATH

Moreover, rpath should be empty in installed shared libs.
Usually, in a default CMake configuration, binaries produced in build tree have absolute paths of dependencies (direct & transitive) libs folder in their rpath, so that you can run your executable for free (on macOS, it only works if external shared libs also have `@rpath` token, so proper relocatable binaries), then it is cleared by CMake during installation.

It also means that all CMake based recipe must:
- use `conan_basic_setup(KEEP_RPATHS)`
- recipe should ensure that https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042 is enabled (either by injection `CMAKE_POLICY_DEFAULT_CMP0042 NEW`, or ensuring `cmake_minimum_required(VERSION 3.0)` or higher in upstream CMakeLists).
- recipe should call install target, not manually copying files (it can be allowed if target os is Windows), so that CMake can:
   - eventually replace absolute path set in shared libs of the build tree with `@rpath` for shared libs in install tree (usually it's already `@rpath` in the build tree, but not always).
   - clear rpath (and it's also important on Linux). To do so, `CMAKE_INSTALL_RPATH` & `CMAKE_INSTALL_RPATH_USE_LINK_PATH` should not be manipulated.
  
   If manual copy can't be avoided, something like this at the end of `package()` may allow to add `@rpath` token into dylib files :
   ```python
    if tools.is_apple_os(self.settings.os):
        with tools.chdir(os.path.join(self.package_folder, "lib")):
            for dylib in glob.glob("*.dylib"):
                command = "install_name_tool -id {0} @rpath/{1}".format(os.path.basename(dylib), dylib)
                self.run(command)
    ```

I suspect that many issues consumers have with "all shared" builds (`-o *:shared=True`) on macOS would be solved. Default behavior of `conan_basic_setup` on macOS is really harmful.

see https://github.com/conan-io/conan-center-index/issues/9052
also https://github.com/conan-io/conan/issues/1238 or https://github.com/conan-io/conan/issues/10253
https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/RunpathDependentLibraries.html
https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#mac-os-x-and-the-rpath
https://www.mikeash.com/pyblog/friday-qa-2009-11-06-linking-and-install-names.html

uilianries:
- [ ] Check KEEP_RPATH on CMakeLists.txt

closes #376

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -74,6 +74,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H065": "NO REQUIRED_CONAN_VERSION",
              "KB-H066": "SHORT_PATHS USAGE",
              "KB-H068": "TEST_TYPE MANAGEMENT",
+             "KB-H071": "RELOCATABLE SHARED",
              }
 
 
@@ -793,6 +794,14 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                     out.error(f"The attribute 'test_type' only should be used with 'explicit' value.: {test_type}")
             except Exception as e:
                 out.warn("Invalid conanfile: {}".format(e))
+
+    @run_test("KB-H071", output)
+    def test(out):
+        cmakefile_path = os.path.join(os.path.dirname(conanfile_path), "CMakeLists.txt")
+        if os.path.isfile(cmakefile_path):
+            cmakefile_content = tools.load(cmakefile_path)
+            if "conan_basic_setup(keep_rpaths)" not in cmakefile_content.lower():
+                out.warn("Did not find 'conan_basic_setup(KEEP_RPATHS)' in CMakeLists.txt. Update your CMakeLists.txt.")
 
 
 @raise_if_error_output
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -74,7 +74,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H065": "NO REQUIRED_CONAN_VERSION",
              "KB-H066": "SHORT_PATHS USAGE",
              "KB-H068": "TEST_TYPE MANAGEMENT",
-             "KB-H071": "RELOCATABLE SHARED",
+             "KB-H071": "RELOCATABLE SHARED LIBS",
              }
 
 
--- a/tests/test_hooks/conan-center/test_relocatable_libs.py
+++ b/tests/test_hooks/conan-center/test_relocatable_libs.py
@@ -1,0 +1,45 @@
+import os
+import textwrap
+
+from conans import tools
+
+from tests.utils.test_cases.conan_client import ConanClientTestCase
+
+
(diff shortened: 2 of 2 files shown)


SpaceIm:
The check should be more powerful: https://github.com/conan-io/hooks/issues/376#issuecomment-1021424535

SSE4:
yes, take a look at https://github.com/conan-io/hooks/pull/171, it inspects dylib/otool output
I think it should validate the binaries itself, not the cmake files (as there are other build systems than cmake, that may potentially suffer exactly that issue - maybe autotools, meson, etc).

SpaceIm:
It's worth noting that I still don't know how to install shared libs with `@rpath` install_name with Meson.

SSE4:
> It's worth noting that I still don't know how to install shared libs with `@rpath` install_name with Meson (currently, all Meson based recipes put absolute path in install_name, like autotools, which is bad).

maybe we can reach to meson developers, in the past, they were friendly enough to help us
but in general, my point that hook should be irrelevant to the build system

SpaceIm:
> but in general, my point that hook should be irrelevant to the build system

Agree, I just want to say that before adding any hook in conan-center, we should ensure that we have a workaround for all build systems, otherwise it may block PR of several recipes.

SSE4:
> Agree, I just want to say that before adding any hook in conan-center, we should ensure that we have a workaround for all build systems, otherwise it may block PR of several recipes.

I worry, this one (and #171) could be enabled only as warnings.
reason - we don't have anything to check `post_package` hook for all the packages, so it's hard to estimate the amount of damage before enabling it. I would personally say, the majority of libraries are still not prepared for that, so we don't want to block new contributors doing PRs (like bump version) forcing them to fix that complex errors.
it's techinically possible to run something that downloads all the packages and runs a hook against a package folder, but it will take hours (if not days) to finish :(

SpaceIm:
Regarding Meson, I asked on Slack but never get any answer: https://cpplang.slack.com/archives/C01AVS654P8/p1643164802012000

This PR seems to have removed any capability to install relocatable shared libs on macOS with Meson: https://github.com/mesonbuild/meson/pull/3691
I don't know whether an option is available to change install_name.

SSE4:
@uilianries conflicts, please rebase

PR #407, opened 2022-04-26
Checksum entries must be filled

### LINKED ISSUE #406: [conan-center] Validate sha content from conandata.yml
Related to conan-io/conan-center-index#9980

The sha256 key has an empty value, which passed by CI without errors. It must be considered an error, as the checksum is mandatory.

uilianries:
The entries `['md5', 'sha1', 'sha256', 'url']`can not be used as an empty key. Otherwise, it will be considered an error.

fixes #406

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -427,14 +427,11 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                         return
             for element in conandata_yml[entry][version]:
                 if entry == "sources" and element in allowed_sources and not conandata_yml[entry][version][element]:
-                    out.error(f"The entry {element} cannot be empty in conandata.yml.")
-                if entry == "sources" and element in weak_checksums:
-                    out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
+                    out.error(f"The entry '{element}' cannot be empty in conandata.yml.")
                 if entry == "sources" and element in checksums:
                     found_checksums.append(element)
-            if found_checksums:
-                if len(found_checksums) > 1 and 'sha256' in found_checksums:
-                    out.warn("Use only 'sha256' as checksum. It's considerably more secure than others.")
+            if found_checksums and 'sha256' not in found_checksums:
+                out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
 
 
     @run_test("KB-H034", output)
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -363,7 +363,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
         allowed_first_level = ["sources", "patches"]
-        allowed_sources = ["url", "sha256", "sha1", "md5"]
+        allowed_sources = ["md5", "sha1", "sha256", "url"]
         allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5"]
 
         def _not_allowed_entries(info, allowed_entries):
@@ -422,6 +422,10 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                     if not validate_recursive(element, conandata_yml[entry][version], "sources",
                                               allowed_sources):
                         return
+            for element in conandata_yml[entry][version]:
+                if entry == "sources" and element in allowed_sources and not conandata_yml[entry][version][element]:
+                    out.error(f"The entries {allowed_sources} must be filled in conandata.yml.")
+
 
     @run_test("KB-H034", output)
     def test(out):
@@ -1488,4 +1492,3 @@ def _load_conanfile(conanfile_path):
                                        python_requires=python_requires,
                                        generator_manager=None)
     return conanfile_obj
-
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -365,6 +365,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
         allowed_first_level = ["sources", "patches"]
         allowed_sources = ["md5", "sha1", "sha256", "url"]
         allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5"]
+        weak_checksums = ["md5", "sha1"]
 
         def _not_allowed_entries(info, allowed_entries):
             not_allowed = []
@@ -425,6 +426,8 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
             for element in conandata_yml[entry][version]:
                 if entry == "sources" and element in allowed_sources and not conandata_yml[entry][version][element]:
                     out.error(f"The entries {allowed_sources} must be filled in conandata.yml.")
+                if entry == "sources" and element in weak_checksums:
+                    out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
 
 
     @run_test("KB-H034", output)
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -366,6 +366,8 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
         allowed_sources = ["md5", "sha1", "sha256", "url"]
         allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5"]
         weak_checksums = ["md5", "sha1"]
+        checksums = ["md5", "sha1", "sha256"]
+        found_checksums = []
 
         def _not_allowed_entries(info, allowed_entries):
             not_allowed = []
@@ -428,6 +430,11 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                     out.error(f"The entries {allowed_sources} must be filled in conandata.yml.")
                 if entry == "sources" and element in weak_checksums:
                     out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
+                if entry == "sources" and element in checksums:
+                    found_checksums.append(element)
+            if found_checksums:
+                if len(found_checksums) > 1 and 'sha256' in found_checksums:
+                    out.warn("Use only 'sha256' as checksum. It's considerably more secure than others.")
 
 
     @run_test("KB-H034", output)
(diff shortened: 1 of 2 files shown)


ericLemanissier:
Maybe we should enforce that there is a sha256 key too?

uilianries:
It could be an warning, but we can't enforce to sah256, because as Croydon mentioned, some projects only provide md5 or sha1.

Croydon:
> It could be an warning, but we can't enforce to sah256, because as Croydon mentioned, some projects only provide md5 or sha1.

We could still enforce it. People can calculate the sha256 checksum as always

uilianries:
> We could still enforce it. People can calculate the sha256 checksum as always

Yes. Checking on Conan Center Index, we have only 3 recipes using non-SHA-256 checksum. Let's use as an warning now, then fix those recipe, and then we can promote to error level.

```
$ find recipes -name conandata.yml | xargs grep sha1  
recipes/soxr/all/conandata.yml:    sha1: "32ea46b1a8c0c15f835422892d02fce8286aec3c"
recipes/uchardet/all/conandata.yml:    sha1: de19b7e614f11572582a0d47f7d5d6f8ec649199
$ find recipes -name conandata.yml | xargs grep md5  
recipes/innoextract/all/conandata.yml:    md5: "964f39bb3f8fd2313629e69ffd3dab9f"
```

danimtb (dismissed):

hooks/conan-center.py
```diff
@@ -422,6 +423,12 @@ def validate_recursive(e, data, name, allowed):
                     if not validate_recursive(element, conandata_yml[entry][version], "sources",
                                               allowed_sources):
                         return
+            for element in conandata_yml[entry][version]:
+                if entry == "sources" and element in allowed_sources and not conandata_yml[entry][version][element]:
+                    out.error(f"The entries {allowed_sources} must be filled in conandata.yml.")
+                if entry == "sources" and element in weak_checksums:
+                    out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
```
jgsogo, line 430: Also, this raises the warning for any _weak checksum_ regardless of `sha256` being included or not.

tests/test_hooks/conan-center/test_conandata.py
```diff
@@ -419,3 +419,30 @@ def test_unknown_subentry_in_list(self):
...
+            conandata = textwrap.dedent(f"""
+                sources:
+                  "1.70.0":
+                    url: "url1.69.0"
+                    {checksum}: "cf23df2207d99a74fbe169e3eba035e633b65d94"
+                """)
+            tools.save('conandata.yml', content=conandata)
+            output = self.conan(['export', '.', 'name/1.70.0@jgsogo/test'])
+            self.assertIn("WARN: [CONANDATA.YML FORMAT (KB-H030)]", output)
+            self.assertIn("Consider 'sha256' instead of ['md5', 'sha1']. It's considerably more secure than others.", output)
```
jgsogo, line 448: Add test with:

```python
    def test_prefer_sha256(self):
        tools.save('conanfile.py', content=self.conanfile)
        for checksum in ['md5', 'sha1']:
            conandata = textwrap.dedent(f"""
                sources:
                  "1.70.0":
                    url: "url1.69.0"
                    sha256: "<sha256>"
                    {checksum}: "cf23df2207d99a74fbe169e3eba035e633b65d94"
                """)
            tools.save('conandata.yml', content=conandata)
            output = self.conan(['export', '.', 'name/1.70.0@jgsogo/test'])
            # TODO: No warnings not errors
```

hooks/conan-center.py
```diff
@@ -422,6 +423,12 @@ def validate_recursive(e, data, name, allowed):
                     if not validate_recursive(element, conandata_yml[entry][version], "sources",
                                               allowed_sources):
                         return
+            for element in conandata_yml[entry][version]:
+                if entry == "sources" and element in allowed_sources and not conandata_yml[entry][version][element]:
+                    out.error(f"The entries {allowed_sources} must be filled in conandata.yml.")
```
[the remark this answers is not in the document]
uilianries: No. It means, the element found in conandata.yml is in `["md5", "sha1", "sha256", "url"]` and it does not a listed value, like `"sha256": ""`, then it's considered an error. Missing keys are not considered, only explicitly declared in conandata.yml.

hooks/conan-center.py
uilianries: True, but we don't have cases where people add more than one checksum. As we want only sha256, let's consider it a prone error.

hooks/conan-center.py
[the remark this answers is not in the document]
danimtb: I think this only checks that if they are declared, they should be filled with something and not empty

tests/test_hooks/conan-center/test_conandata.py
uilianries: Done. Test `test_sha1_md5` already does it. Just updated it.

hooks/conan-center.py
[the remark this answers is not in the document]
jgsogo:
```suggestion
                    out.error(f"The entry {element} cannot be empty in conandata.yml.")
```

tests/test_hooks/conan-center/test_conandata.py
```diff
@@ -262,6 +262,8 @@ def test_sha1_md5(self):
         tools.save('conandata.yml', content=conandata)
         export_output = self.conan(['export', '.', 'name/1.69.0@jgsogo/test'])
         self.assertNotIn("ERROR: [CONANDATA.YML FORMAT (KB-H030)]", export_output)
+        self.assertIn("WARN: [CONANDATA.YML FORMAT (KB-H030)]", export_output)
+        self.assertIn("Use only 'sha256' as checksum. It's considerably more secure than others.", export_output)
```
jgsogo, line 266: I don't think we shouldn't recommend to use **only** `sha256` if others are already there (Artifactory uses the three, and we are also JFrog).

tests/test_hooks/conan-center/test_conandata.py
jgsogo: Yes, that's it. In that test `sha256` is set, so the message is misleading. It is ok to use `sha256` (required) AND others.

tests/test_hooks/conan-center/test_conandata.py
uilianries: Okay. Checking Conan sources, it validates all 3, so it does not matter which one is used besides sha256: https://github.com/conan-io/conan/blob/8ce1c508d50584886bce76dd0c14e337499c94e5/conans/client/downloaders/file_downloader.py#L18 Let's keep it as normal usage and remove the warning

uilianries: Done

hooks/conan-center.py
uilianries: Done. Now it raises a warning only when sha256 is not listed

hooks/conan-center.py
```diff
@@ -363,8 +363,11 @@ def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
         allowed_first_level = ["sources", "patches"]
-        allowed_sources = ["url", "sha256", "sha1", "md5"]
+        allowed_sources = ["md5", "sha1", "sha256", "url"]
```
uilianries, line 366: keep alphabetical order due Windows behavior when printing out a list on terminal.

hooks/conan-center.py
```diff
@@ -427,14 +427,11 @@ def validate_recursive(e, data, name, allowed):
...
-                if entry == "sources" and element in weak_checksums:
-                    out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
+                    out.error(f"The entry '{element}' cannot be empty in conandata.yml.")
                 if entry == "sources" and element in checksums:
                     found_checksums.append(element)
-            if found_checksums:
-                if len(found_checksums) > 1 and 'sha256' in found_checksums:
-                    out.warn("Use only 'sha256' as checksum. It's considerably more secure than others.")
+            if found_checksums and 'sha256' not in found_checksums:
+                out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
```
jgsogo, line 434: This check should be..., shouldn't it?

```python
if not found_checksums or 'sha256' not in found_checksums:
    out.error(f"sha256 checksum is required")
```

uilianries: looking the code, indeed there is no check for it! Let me update it. However, we can consider both an error, only the case where is empty. Because we have some recipes in Conan Center using another checksum that's not sha256.

jgsogo: We can check them (we can run `ScheduleExportCheck` using this branch) and see if we can afford fixing them. If they are 10-20 we can create those PRs during this week.

hooks/conan-center.py
```diff
@@ -422,6 +445,15 @@ def validate_recursive(e, data, name, allowed):
...
                                               allowed_sources):
                         return
+            for element in conandata_yml[entry][version]:
+                if entry == "sources":
+                    has_sources = True
+                    validate_checksum_recursive(element, conandata_yml[entry][version])
+            if not found_checksums and has_sources:
+                out.error("The checksum key 'sha256' must be declared and can not be empty.")
+            elif found_checksums and 'sha256' not in found_checksums:
+                out.warn(f"Consider 'sha256' instead of {weak_checksums}. It's considerably more secure than others.")
```
jgsogo, line 455: IMHO, this should be and ERROR

jgsogo (approved):
If we compare with the latest execution from `master` branch, there are 40 new errors: https://github.com/conan-io/conan-center-index/issues/2232 Not so many, I'm fine merging it, they can be fixed as we keep getting new pull-requests.

hooks/conan-center.py
uilianries: I agree, but we can't do it now. First we need to fix those recipes.

jgsogo: We don't know how many recipes would fail if that were an ERROR... Change it to error here, then run `ScheduledExportCheck` (reporting results here) and we can decide depending on the number of recipes that fail.

uilianries: The most important case is missing checksum cases. We usually enforce users to use sha-256 when reviewing

jgsogo: Ok, no problem. We can merge this one and open another PR with the ERROR and see which are those recipes.

hooks/conan-center.py
SSE4: what does it mean?

SSE4 (approved):

hooks/conan-center.py
uilianries: When I run a test with a list on Linux, it does not change anything, but for some reason, using tox on windows, it forces a sort over the list, so if the order matters, it will fail on windows.

Croydon (approved):

danimtb (approved):

PR #410, opened 2022-05-03
[conan-center] KB-H030 need to accept a new edge case

uilianries:
The official recipe zulu-openjdk uses pre-built binaries which is not covered by the current implementation of conan-center hook.

Related to https://github.com/conan-io/conan-center-index/pull/10592#issuecomment-1115887260

**NOTE**: For some reason mac was failing due max path length. I needed to change the test folder path to `/tmp/py39` (example) and it works now.


uilianries:
**NOTE**: For some reason mac was failing due max path length. I needed to change the test folder path to `/tmp/py39` (example) and it works now.

SSE4 (approved):

jgsogo (approved):

PR #411, opened 2022-05-05
[conan-center] Do not allow build_policy attribute

uilianries:
The build policy on Conan Center should be only managed by the CI service. Users should not force any custom build policy.

So far, we don't have any recipe in CCI with build_policy.


SSE4 (approved):
easy!

jgsogo (approved):

PR #412, opened 2022-05-05
[conan-center] create-dmg should be accepted as header-only

uilianries:
That project only exports a single file which is a script. It does matters any setting.

/cc @MartinDelille

Related to https://github.com/conan-io/conan-center-index/pull/10382


SSE4 (approved):

SSE4:
this is just a shell script: https://github.com/create-dmg/create-dmg/blob/master/create-dmg

PR #416, opened 2022-05-09
Use Python 3.7 for Mac

uilianries:
- Mac is failing to install Python 3.9
- We don't use Python 3.9 on CCI, let's use Python 3.7

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -37,50 +37,50 @@ pipeline {
                         }
                     }
                 }
-                stage('Macos - py39') {
+                stage('Macos - py37') {
                     agent {
                         label 'Macos'
                     }
                     stages {
-                        stage('Macos - py39 - Generate environment') {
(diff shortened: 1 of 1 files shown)


SSE4 (dismissed):

jgsogo:
Hooks are more than ConanCenter, so better to move to 3.10 than move back to 3.7. I would use `3.7` (the one in ConanCenter) and the latest python version (likely the one used by users).

uilianries:
> Hooks are more than ConanCenter, so better to move to 3.10 than move back to 3.7. I would use `3.7` (the one in ConanCenter) and the latest python version (likely the one used by users).

The idea is good, but we don't have Python 3.10 installed on Mac.

uilianries:
@SSE4 It's working with python 3.8 now. Let's switch to 3.8 to fix it, then we can think about 3.10 on the future.

SSE4 (approved):

PR #418, opened 2022-05-12
add hook checking include path existence

ericLemanissier:
avoids https://github.com/conan-io/conan-center-index/pull/10743#issuecomment-1125033187


uilianries (approved):

SSE4 (approved):

uilianries:
It looks correct, based on previous hooks, but I would ask a test here, as it is classified as error, so if something goes wrong, it will break CCI for awhile

ericLemanissier:
mmm `ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/private/var/folders/c2/vd5y5mfs33q84hyzmxvqmyjm0000gp/T/py38/lib/python3.8/site-packages/platformdirs-2.5.2.dist-info/METADATA'`
That's not on me right?

uilianries:
@ericLemanissier no, we had this same error few week ago, but only when running Python 3.9

ericLemanissier:
It's ready for review @SSE4 @uilianries

uilianries (approved):
Amazing!

danimtb (approved):

PR #425, opened 2022-06-17
ignore global cpp_info when there are components

ericLemanissier:
KB-H054 and KB-H071
unblocks conan-io/conan-center-index#10795


danimtb (approved):
Thanks for the fix! I was looking into the error as well and I thought the message was wrong but your fix is the good one

uilianries (approved):

uilianries:
Need to update the tests. At least this time is not another infra error.

SSE4 (approved):

PR #428, opened 2022-06-22
[fix] Pin pylint version to overcome CI error

danimtb:
Unblocking the latest PRs like #426 or specifically #425

### DIFF
--- a/tests/requirements_test.txt
+++ b/tests/requirements_test.txt
@@ -2,7 +2,7 @@ pytest>=3.6
 parameterized
 responses
 pluggy
-pylint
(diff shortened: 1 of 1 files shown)


ericLemanissier:
A more sustainable fix would be to detect corrupted environments and remove their folder automatically, or don't cache any environment in the first place

danimtb:
Thanks for the hint folks! It is green now 😄 let's see if we can unblock some PRs 🤾‍♂️

ericLemanissier (approved):

uilianries (approved):

PR #429, opened 2022-06-22
- conandata.yml reduce hook for 2.0

SSE4:
generates the same RREV as v1:
```
fmt/8.1.1: Exported revision: f0c7ec2a07a4a1ba80bfc2e4cbf18a43

f0c7ec2a07a4a1ba80bfc2e4cbf18a43
```
with v2:
```
←[1m←[32mExported recipe: fmt/8.1.1#f0c7ec2a07a4a1ba80bfc2e4cbf18a43 (2022-06-22 11:12:33 UTC)←[0m 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Repository to develop **experimental** [Conan](https://conan.io) hooks for Conan
  * [Recipe linter](#recipe-linter)
  * [Non ASCII](#non-ascii)
  * [YAML linter](#yaml-linter)
+ * [Reduce conandata.yml](#reduce-conandata)
 
 
 ## Hook setup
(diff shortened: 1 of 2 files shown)


SSE4:
> Is it not the same as "KB-H031": "CONANDATA.YML REDUCE"?

it is the same, plus some minor changes for 2.0 api.
right now it's faster/easier to extract it to its own hook, instead of migrating the whole file (which I doubt we will need, as most of checks could be moved to linter).

SSE4:
> @SSE4 sounds reasonable. Please, update the README too.

what exactly?

uilianries (approved):

danimtb:
I think it is fine to have it separated from the rest of hooks. This way we can clearly know what hooks conan v2 would use.
In the mid term, as we start to add some more hooks here, we would have to add tests for them as well.

danimtb (approved):

PR #437, opened 2022-07-18
[v2] Put v2 hook in quarantine

jgsogo:
Deactivate hook until we can really fix it


uilianries:
The platformdirs error strikes again! It was the test folder in the past.

SSE4 (approved):

ericLemanissier:
It's the usual `ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/private/var/folders/c2/vd5y5mfs33q84hyzmxvqmyjm0000gp/T/3/py38/lib/python3.8/site-packages/platformdirs-2.5.2.dist-info/METADATA'` https://ci.conan.io/blue/organizations/jenkins/Hooks/detail/PR-437/3/pipeline
which happens because the build folder is corrupted, and never cleaned up.

SSE4:
unfortunately, this is beyond my python-fu knowledge :(

ericLemanissier:
it's not a python problem, but a CI problem. The build folder should be completely erased after a job has been interrupted, or after a job failed. folder `/private/var/folders/c2/vd5y5mfs33q84hyzmxvqmyjm0000gp/T/1`, `/private/var/folders/c2/vd5y5mfs33q84hyzmxvqmyjm0000gp/T/2` and `/private/var/folders/c2/vd5y5mfs33q84hyzmxvqmyjm0000gp/T/3` have a corrupted pyenv.
A more radical approach would be to completely clear the pyenv folder before starting the job.

jgsogo:
I can add this to Jenkins configuration:

**Clean before checkout**

> Clean up the workspace before every checkout by deleting all untracked files and directories, including those which are specified in .gitignore. It also resets all tracked files to their versioned state. This ensures that the workspace is in the same state as if you cloned and checked out in a brand-new empty directory, and ensures that your build is not affected by the files generated by the previous build.

**Wipe out repository & force clone**
> Delete the contents of the workspace before building, ensuring a fully fresh workspace.

Maybe it helps

ericLemanissier:
https://github.com/conan-io/hooks/pull/438 should be the fix

uilianries (approved):
LGTM

PR #438, opened 2022-07-18
fix macos CI

ericLemanissier:


.ci/generate_env_macos.sh
```diff
@@ -21,6 +21,7 @@ case "${PYVER}" in
         ;;
 esac
 
+rm -r ${TEST_FOLDER}
```
jgsogo, line 24: IMO it would be better to put this folder inside the Jenkins' workspace... if needed, we can pass the argument from the Jenkinsfile.

Is there any reason not to do it? cc/ @uilianries

jgsogo (approved):

.ci/generate_env_macos.sh
ericLemanissier: I agree

jgsogo: This is not going to work, there are two calls to `generate_env_macos` in parallel, so this can lead to unexpected errors.

 * https://github.com/conan-io/hooks/blob/master/.ci/Jenkinsfile#L17
 * https://github.com/conan-io/hooks/blob/master/.ci/Jenkinsfile#L50
 
We really need to pass an argument from the Jenkinsfile: environment is already available, so it should be straightforward to get here the `WORKSPACE`. This should work, and there is no need to remove the folder:

```
TEST_FOLDER="${WORKSPACE}/${PYVER}"
```

jgsogo (changes requested):

.ci/generate_env_macos.sh
ericLemanissier: For the record, this should not have conflicted because these two invocations have a different `PYVER` env var values, and this variable is part of `TEST_FOLDER`.
I did the change anyway because it is cleaner.

jgsogo: You are right! 👍

jgsogo (approved):

Issue #440, opened 2022-07-27
[conan-center] Prefer built in tools over recipe methods


prince-chrismc:
Conan v1.45 got `tools.microsoft.is_msvc` and this was extensively implemented by hand in a lot of recipes.
https://github.com/conan-io/conan-center-index/search?q=def+_is_msvc gives 141 results.

I would be nice to start encouraging contributors to update.

There would be other msvc tools like "static runtime" but I am not sure how that could work https://github.com/conan-io/conan-center-index/search?q=mt+in+compiler+runtime or https://github.com/conan-io/conan-center-index/search?q=mt+in+self+settings give very different results

uilianries:
@prince-chrismc Thank you for your suggestion, indeed it's a valid topic, it will be moved to CCI linter, as it's more a static validation and linter has a better transparency when reviewing.

PR #441, opened 2022-07-28
[conan-center] The custom method _is_msvc is no longer needed

uilianries:
New Conan Center hook entry: 72 - MSVC TOOLS

The first step is suggesting the deprecation of `_is_msvc`.

Related to #440 

/cc @prince-chrismc 

WIKI INFO:

**#KB-H072: "MICROSOFT TOOLS"**

Some custom and popular methods involving Microsoft compiler has been adopted to Conan client. Use only methods provided by `conan.tools.microsoft` import instead of adopting custom methods, they are better maintained and easier to update.


uilianries:
It should be moved to CCI linter.

PR #442, opened 2022-08-04
[conan-center] Accept self.info.clear() as header-only

uilianries:
According to the migration guide, the `self.info.clear()` should be used to identify header-only recipes.

Documentation https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html#the-package-id-method

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1357,7 +1357,8 @@ def _is_recipe_header_only(conanfile):
     package_id_method = getattr(conanfile, "package_id")
     header_only_id = "self.info.header_only()" in inspect.getsource(package_id_method)
     settings_clear = "self.info.settings.clear()" in inspect.getsource(package_id_method)
-    return header_only_id or without_settings or settings_clear
+    info_clear = "self.info.clear()" in inspect.getsource(package_id_method)
+    return header_only_id or without_settings or settings_clear or info_clear
 
 
 def _get_settings(conanfile):


danimtb:
Moreover, we should add a migration check to the linter to use `self.info.clear()` instead of `self.info.header_only()`

uilianries:
@danimtb Yes, but still, this change id needed anyway, otherwise, conan-center hook complains about `self.info.clear()` usage.

danimtb (approved):

danimtb:
@uilianries do you have an example of that? did it happen already?

uilianries:
@danimtb Yes: https://github.com/conan-io/conan-center-index/pull/11981#issuecomment-1205022505

I needed to revert to `self.info.header_only` because Hooks didn't recognized my recipe as header-only and complained that was missing artifacts.

jgsogo:
So `self.info.header_only()` is something that should never be used? The linter can just remove it from `info` object and syntax will complain

uilianries:
> So `self.info.header_only()` is something that should never be used? The linter can just remove it from `info` object and syntax will complain

As soon as we run Conan 1.50 on CCI, we can apply that rule with the linter.

jgsogo:
We are already running `1.50.1`, so it's something we can already add.

I see in Conan codebase:

```python
class ConanInfo(object):
    ...

    # FIXME: Rename this to "clear" in 2.0
    def header_only(self):
        self.settings.clear()
        self.options.clear()
        self.requires.clear()

    clear = header_only
```

uilianries:
> We are already running 1.50.1, so it's something we can already add.

Yes, just in case you need to revert ... 😏 

Anyway, I'm using `self.info.clear()` [here](https://github.com/conan-io/conan-center-index/pull/11981/files#diff-01c43ca6e9b97cfb75be1640171375906433f05b6f426fe27581328967b669b8R70) and Conan Center, including linter, accepted.

PR #449, opened 2022-09-20
[conan-center] Pylint should not be skipped from conanfile

uilianries:
Based on CCI activity. As humans, we can fail when reviewing it, pylint will skip silently.

Wiki:

#### **<a name="KB-H072">#KB-H072</a>: "PYLINT EXECUTION"**

Pylint is executed by default over all `conanfile.py` in Conan Center Index and it should not be skipped. It's an important tool which help us keeping a standard level of acceptance, otherwise, it would be incredible hard to review all recipes keeping the same standard level.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -883,13 +883,11 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
                 if re.search(pattern, content):
                     out.error(f"Pylint can not be skipped, remove '#pylint' line from '{path}'")
 
-        _check_conanfile_content(conanfile_content, "conanfile.py")
-        test_package_path = os.path.join(os.path.dirname(conanfile_path), "test_package", "conanfile.py")
-        test_v1_package_path = os.path.join(os.path.dirname(conanfile_path), "test_v1_package", "conanfile.py")
-        for test_package in [test_package_path, test_v1_package_path]:
-            if os.path.exists(test_package):
-                test_package_content = tools.load(test_package)
-                _check_conanfile_content(test_package_content, test_package)
+        recipe_folder = os.path.dirname(conanfile_path)
+        recipes = _get_files_following_patterns(recipe_folder, "conanfile*.py")
+        for recipe in recipes:
+            recipe_content = tools.load(recipe)
+            _check_conanfile_content(recipe_content, recipe)
 
 
 @raise_if_error_output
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -874,6 +875,22 @@ def test(out):
...
+        def _check_conanfile_content(content, path):
+            patterns = [r'#\s*pylint\s*:\s*skip-file\s*', '#\s*pylint\s*:\s*disable-all\s*', '#\s*pylint\s*:\s*disable=']
+            for pattern in patterns:
+                if re.search(pattern, content):
+                    out.error(f"Pylint can not be skipped, remove '#pylint' line from '{path}'")
+
+        _check_conanfile_content(conanfile_content, "conanfile.py")
+        test_package_path = os.path.join(os.path.dirname(conanfile_path), "test_package", "conanfile.py")
+        test_v1_package_path = os.path.join(os.path.dirname(conanfile_path), "test_v1_package", "conanfile.py")
+        for test_package in [test_package_path, test_v1_package_path]:
```
jgsogo, line 889: This could be a little bit more generic, and check any `conanfile.py` inside the directory (there might be many `test_xxxx` folders)

jgsogo (dismissed):
It can be generalized later if needed

hooks/conan-center.py
uilianries: True, I'll update it.

uilianries: Done.

hooks/conan-center.py
```diff
@@ -874,6 +875,20 @@ def test(out):
...
+    @run_test("KB-H072", output)
+    def test(out):
+        def _check_conanfile_content(content, path):
+            patterns = [r'#\s*pylint\s*:\s*skip-file\s*', '#\s*pylint\s*:\s*disable-all\s*', '#\s*pylint\s*:\s*disable=']
+            for pattern in patterns:
+                if re.search(pattern, content):
+                    out.error(f"Pylint can not be skipped, remove '#pylint' line from '{path}'")
+
+        recipe_folder = os.path.dirname(conanfile_path)
+        recipes = _get_files_following_patterns(recipe_folder, ["conanfile*.py"])
```
jgsogo, line 887:
```suggestion
        recipes = _get_files_following_patterns(recipe_folder, ["conanfile.py", ])
```

why the `*`? Is it regex (better `\.`)?

jgsogo (approved):
Looks good. Let's see it running live

danimtb (approved):

PR #451, opened 2022-09-30
[conan-center] KB-H073 Test v1 package folder

### LINKED ISSUE #450: [conan-center] When test_package uses new CMake then test_v1_package must be present
To keep Conan 1.x compatibility test, any package which provides test_package using conan.tools.cmake.CMake, should provide test_v1_package too.

uilianries:
As users are porting Conan v2 package in CCI, but mostly forget to add test_v1_package, this hook should help them:

- When test_package is ported to support Conan v2, test_v1_package must be present
- test_v1_package must contain a conanfile.py
- It will be considered an error

closes #450 

#### **<a name="KB-H073">#KB-H073</a>: "TEST V1 PACKAGE FOLDER"**

The legacy content in test_package should not be removed, instead, rename that folder to test_v1_package and create a new test_folder following the [file structure](https://github.com/conan-io/conan-center-index/blob/master/docs/how_to_add_packages.md#recipe-files-structure) related to Conan v2 and v1 compatibility. Also, you can obtain good example on [templates package](https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates) about Conan v2 test_package.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -82,6 +82,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H070": "MANDATORY SETTINGS",
              "KB-H071": "INCLUDE PATH DOES NOT EXIST",
              "KB-H072": "PYLINT EXECUTION",
+             "KB-H073": "TEST V1 PACKAGE FOLDER",
              }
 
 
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -890,6 +891,24 @@ def _check_conanfile_content(content, path):
...
+        test_package_conanfile_path = os.path.join(dir_path, "test_package", "conanfile.py")
+        test_v1_package_path = os.path.join(dir_path, "test_v1_package")
+        test_v1_package_conanfile_path = os.path.join(test_v1_package_path, "conanfile.py")
+        content = None
+        if os.path.isfile(test_package_conanfile_path):
+            try:
+                content = tools.load(test_package_conanfile_path)
+            except Exception as error:
+                out.warn(f"Invalid conanfile: {error}")
+        if content and ("VirtualRunEnv" in content or "self.cpp.build" in content):
```
danimtb, line 906: I think checking for `self.requires(self.tested_reference_str)` or just `self.tested_reference_str` should be enough, as explicit requirement is mandatory in conan v2 and that one is totally missing in all test_package (v1) recipes. WDYT?

uilianries, line 906:
```suggestion
        if content and "self.tested_reference_str" in content:
```

uilianries: Done. Yes, it makes more sense.

danimtb (approved):

PR #455, opened 2022-11-15
[conan-center] KB-H074 - check if there is at least one static lib when shared=False

SpaceIm:
Migration to CMakeToolchain may break shared=False when upstream CMakeLists overrides BUILD_SHARED_LIBS as an option, or a CACHE variable. While https://github.com/conan-io/conan/pull/12401 may avoid most issues in conan >= 1.54.0, there might be some cases where it might still fail (like https://github.com/conan-io/conan-center-index/issues/14206).

So I propose this hook to check that `shared=False` is honored. It has the same logic than KB-H015 (check of shared artifacts). It's worth noting that check may not detect that static libs were not created if compiler is Visual Studio since static & import libs use the same extensions.

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -83,6 +83,7 @@ kb_errors = {"KB-H001": "DEPRECATED GLOBAL CPPSTD",
              "KB-H071": "INCLUDE PATH DOES NOT EXIST",
              "KB-H072": "PYLINT EXECUTION",
              "KB-H073": "TEST V1 PACKAGE FOLDER",
+             "KB-H074": "STATIC ARTIFACTS",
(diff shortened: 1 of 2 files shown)


AndreyMlashkin (approved):

SpaceIm:
@uilianries @danimtb could you review please? It would be nice to have this check in conan-center ASAP to avoid to produce shared libs when shared=True, given that recipes always consume dependencies with shared=False

uilianries (approved):
LGTM

uilianries:
@SpaceIm merged, merci!

garethsb:
Nice 👍

PR #457, opened 2022-11-22
Skip post package for static library when building header only

### LINKED ISSUE #456: [conan-center] KB-H0074: False positove when using self.info.clear() as optional
The PR https://github.com/conan-io/conan-center-index/pull/14330 illustrates a case where a package can be or not header-only and when using `self.info.clear()` to set the package ID, the Conan Center hook does not recognize it properly and expects a static library installed.

/cc @toge

uilianries:
- Post package for header-only is already validated. No artifacts is expected
- When is a header-only package, ignores shared=False on post package.

fixes #456 

/cc @toge

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -1340,10 +1340,13 @@ def _shared_files_well_managed(conanfile, folder):
 def _static_files_well_managed(conanfile, folder):
     static_extensions = ["a", "lib"]
     shared_name = "shared"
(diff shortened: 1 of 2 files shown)


prince-chrismc (approved):

danimtb (approved):

SpaceIm:
Should this logic also be added to `_shared_files_well_managed()`?

PR #463, opened 2022-12-19
White list recipes with proxy version in KB-H010/KB-H012/KB-H014

SpaceIm:
If we agree on version of https://github.com/conan-io/conan-center-index/pull/14814


SSE4:
what's the common agreement on virtual packages so far?

PR #474, opened 2023-01-06
Update kb-h075 to only match active self.requires(..., override=True) statements

### LINKED ISSUE #473: conan-center hook KB-H075 has a false positive when giving example in comments about what consumer should do
The hooks have a bug, it's tripping on https://github.com/conan-io/conan-center-index/blob/a859c2bdd26f68494cfde69bac1061b979bfc368/recipes/armadillo/all/conanfile.py#L182 which matches the regex from what I can figure out

```diff
- # self.requires("intel-mkl/version@user/channel, override=True) in their consumer 
+ # set `override=True` when calling `self.requires` which specifies `intel-mkl/version@user/channel` in their consumer
```

should be enough to work around it

_Originally posted by @prince-chrismc in https://github.com/conan-io/conan-center-index/issues/15109#issuecomment-1372841368_

samuel-emrys:
Fix hook for KB-H075 to ensure that it only matches self.requires statements that are active code. This ensures that it doesn't match similar strings that appear in comments, which may be common when providing advice to consumers.

This prevents the following from being matched:

```python
def requirements(self):
    # self.requires("pkg/version@user/channel", override=True)
```
by ensuring that only lines with only preceeding whitespace are matched. In this case, the following would be matched:

```python
def requirements(self):
    self.requires("pkg/version@user/channel", override=True) # anything here won't affect match
```

Closes #473

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -913,7 +913,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H075", output)
     def test(out):
-        match = re.search(r'self.requires\(.*override=True.*\)', conanfile_content)
+        match = re.search(r'(^|\n)\s*self.requires\(.*override=True.*\)', conanfile_content)
         if match:
             out.error("self.requires('package/version', override=True) is forbidden, do not force override parameter.")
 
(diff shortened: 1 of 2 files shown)


uilianries (approved):
LGTM

uilianries:
@samuel-emrys Thank you for your contribution!!

PR #476, opened 2023-01-15
[conan-center] New KB-H076 - test that static & shared artifacts of the same lib are not packaged together

SpaceIm:
This new hook checks whether static & shared artifacts of the same lib are packaged together.
It checks if there are tuples of (.a/.dylib) or (.a/.so) files with the same name. So it works on *nix systems only, not Windows. Hopefully it should catch most issues.

I've also added explicit tests for the two other hooks checking shared & static artifacts. They were not properly tested.

(by the way it's really tedious to have to use unittests instead of pytest, it's almost impossible to deeply parametrize tests since unittests & pytest quickly break when mixed-up. `parameterized` lib is not as powerful as `pytest.mark.parametrize` which can easily be stacked to test matrices of values).

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -2,6 +2,7 @@ import ast
 import collections
 
 import fnmatch
+import glob
 import inspect
 import os
 import re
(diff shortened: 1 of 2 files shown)


uilianries:
Nice, this new hook complements KB-H015 (shared artifacts only) and KB-H074 (static artifacts only).

uilianries (dismissed):
There are cases where projects names their libraries according to their type e.g libfoo_static.a libfoo.so which would not be matched in KB-H076. However, as first step it looks good enough, most of cases will be covered.

SpaceIm:
> There are cases where projects names their libraries according to their type e.g libfoo_static.a libfoo.so which would not be matched in KB-H076

Yes, it's very hard to detect this case without code introspection. There is a comment in source code about cases not covered by this hook.

uilianries (approved):

PR #483, opened 2023-02-08
Fix Hook url for Conan Center

RubenRBS:
The hooks were generating URLs with only the hook id set as the anchor which to scroll in the document, but the ID of the element also contains the hook title https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#kb-h076-either-static-or-shared-of-each-lib


uilianries (approved):
LGTM

PR #485, opened 2023-02-14
conan v2 hooks: fix self.run usage

SpaceIm:
/cc @uilianries


SpaceIm:
I'm quite concerned that these v2 hooks have no dedicated tested to the same quality level (or even not at all) than v1 hooks.

uilianries (approved):

uilianries:
@SpaceIm Thank you for providing this hotfix. The idea is removing most of hooks which are not focused on Conan v2. It should be a next step actually. Need to check why this hooks is running by default .

PR #489, opened 2023-03-27
[docs] Document conan center V2 status

uilianries:
So far we didn't decide what to do with hooks and Conan v2, that's why we delayed new hooks and linters. We still need to think about, because command extension could be an alternative to be associated too.

This PR documents the situation a bit, so users should not open new issues about Conan v2 not work with Conan center hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,13 @@ variable ``CONAN_HOOK_ERROR_LEVEL``:
    - ``CONAN_HOOK_ERROR_LEVEL=40`` it will raise if any error happen.
    - ``CONAN_HOOK_ERROR_LEVEL=30`` it will raise if any error or warning happen.
 
+#### Conan 2.x support
+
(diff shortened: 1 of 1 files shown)


n-bes:
Just in time 👌

SSE4 (approved):

danimtb (approved):

SSE4 (approved):

SpaceIm:
Any update? What's the status of hooks in conan v2? Do they have to be moved to `conan-extensions`? I would love to run some hooks locally, and several hooks checking binaries or package content can't be delegated to linters.

danimtb:
I think we forgot to merge this. @SpaceIm thanks fot the ping. We have to revisit the hooks in conan v2 and try to do some clean up around them and the linters in conan-center-index.

SpaceIm:
I've coded a conan-extension command checking relocability of macOS dylib files (KB-H077) in local cache, but I still don't know whether cci hooks will be replaced by conan-extension.

PR #501, opened 2023-06-14
Add *.cuh as a file type for CUDA header-only packages

valgur:
Fixes a build failure due to the hook not finding any `*.h` files in a pure CUDA header-only package in this PR: https://github.com/conan-io/conan-center-index/pull/17484


memsharded (approved):
CUDA SDK seems to be using this extension, so looks good: https://stackoverflow.com/questions/2393418/cuda-source-files-get-a-cu-extension-what-do-header-files-get

uilianries (approved):
LGTM

PR #525, opened 2024-01-13
conan-center.py: add an exception for `tz`

valgur:
To fix the invalid linter error in https://github.com/conan-io/conan-center-index/pull/21671
```
[HOOK - conan-center.py] post_package(): ERROR: [MATCHING CONFIGURATION (KB-H014)] Package for Header Only does not contain artifacts with these extensions: ['h', 'h++', 'hh', 'hxx', 'hpp', 'cuh'] (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H014-MATCHING-CONFIGURATION) 
[HOOK - conan-center.py] post_package(): ERROR: [MATCHING CONFIGURATION (KB-H014)] Packaged artifacts does not match the settings used: os=Linux, compiler=gcc (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H014-MATCHING-CONFIGURATION) 
```

PR #526, opened 2024-01-25
[CI] Fix Jenkins pipeline for testing hooks

danimtb:
From PR #525 we realized that the pipeline was getting stuck due to macos machines not available. I thought this was an easy fix but it required to almost completely redo the whole pipeline.

Here is the summary of tasks covered in this PR:
- Remove `tox` usage: It was causing a lot of installation problems and delays in the CI (windows machines even got stuck executing tox).
- Remove Python script to get the latest version of Conan: It was causing some trouble getting the latest Conan v2 instead of the latest v1. Now this is done via pip install.
- Update the Macos agent label to use the new Macos M2 machine.
- Fix some requirement versions creating conflicts.
- Add Conan v2 test for the "reduce conandata" hook.
- Added environments to test with both conan v1 and v2 in develop/develop2, latest and prev (latest -1) versions.
- Continue to test hooks with different Python versions (this might look like an overkill but decided to keep it for the moment and avoid any kind of regression).
- Add script to launch tests in Windows.
- Keep ``hook_conan-center-v2`` hook untested for now.

Testing on Windows is slower than in other environments, that's why I decided to reduce the Python versions tested on this platform.
The last stage in Linux is the complete one with all the combinations: python versions * conanv1/v2 * conan versions (develop, latest, prev).

-----------------------
For future PRs:
- Agree on the Python versions to run tests (align here with the Conan client and its CI)
- Add conanv1prev and conanv2prev versions as global variables in the Jenkinsfile for easy updates (these would be the only things to update in the CI whenever there are new releases).


RubenRBS (dismissed):

uilianries:
Yes, from time to time, the CI gets wild in hooks.

RubenRBS (approved):

RubenRBS (approved):
Approving but waiting for @uilianries last look

.ci/Jenkinsfile
```diff
@@ -6,66 +6,80 @@ pipeline {
...
+                                PYTEST_ARGS = "./tests/test_hooks/conan-center-v2"
                             }
                             steps {
-                                sh '.ci/run_tox.sh'
+                                sh '.ci/generate_env_macos.sh'
+                                sh '.ci/run_tests.sh'
                             }
                         }
-                        stage('Macos - py38 - conancurrent') {
+                        stage('Macos - py38 - conan v2 prev') {
```
uilianries, line 74: I would skip prev because we don't update hook so frequently now and it would need manual update. Probably, we will forget to update it for each new Conan release. In CCI we ask in the PR to users run latest Conan version, so having develop and latest only is good enough to me.

.ci/run_tests.bat
```diff
@@ -0,0 +1,4 @@
+REM https://github.com/conan-io/conan_ci_jenkins/blob/master/resources/org/jfrog/conanci/python_runner/conf.py
+set TEST_FOLDER=D:/J/t/Hooks/%BUILD_NUMBER%/%PYVER%
```
uilianries, line 2: Is there a variable environment instead? Like `%WORKSPACE%` maybe.

.ci/run_tests.sh
```diff
@@ -20,4 +20,4 @@ esac
 
 source ${TEST_FOLDER}/bin/activate
 
-${PYVER} -m tox --recreate
+python -m pytest ${PYTEST_ARGS}
```
uilianries, line 23: Does it really run the correct python version? Usually we had problems by running the system python instead. Maybe adding `python --version` just in case to make it explicit.

uilianries (commented):
Looking good, it's always a pain updating and fix hooks tests.

.ci/run_tests.sh
danimtb: will add it to verify

.ci/run_tests.bat
danimtb: I will try it as well, but I just copy pasted this

.ci/run_tests.sh
danimtb: done. Also added conan version

.ci/run_tests.bat
danimtb: it works!

uilianries (approved):
LGTM.

PR #529, opened 2024-03-08
[bugfix] Shared libraries in Windows (`.ddl.lib` not found as interface)

franramirez688:


RubenRBS (approved):

franramirez688:
Update: The bug comes from `collect_libs` as it's gathering DLL extensions, and it's a bit difficult to fix without breaking. Proposing another way around: https://github.com/conan-io/conan/pull/15866

PR #530, opened 2024-03-21
Remove `conandata.yml` field limits

RubenRBS:
This will allow to have an extra fields in the `conandata.yml` file, which is meant to contain things that the recipes use on a per version basis.
ie, `wasmtime-cpp` (Among many others) contains a version mapping inside the recipe, so every time a new version is added, the revisions of every other version also get modified (https://github.com/conan-io/conan-center-index/pull/23172/files#diff-1a9d4053acc279660d5b2b73ca7f473b969b33b25621bdad9465c9da9f84fd65R47)

This allows the `trim_conandata` hook to keep the old revisions for older versions :)

Note that we'll be conservative when allowing these fields in PR reviews, so a good reason might need to be provided in order for us to accept these if its usage falls outside the example shown below

The idea is to move those kinds of mappings
from
```py
version_map = {
            "0.35.0": "0.35.1",
            "0.39.0": "0.39.1",
            "1.0.0": "1.0.1",
            "6.0.0": "6.0.1",
            "9.0.0": "12.0.2",
            "18.0.0": "18.0.3",
        }
selected = version_map[self.version]
```
to
conandata.yml
```yml
"extra":
   "0.35.0":
      "requirement": "0.35.1"
   "0.39.0":
      "requirement": "0.39.1"
   "1.0.0":
      "requirement": "1.0.1"
   "6.0.0":
      "requirement": "6.0.1"
   "9.0.0":
      "requirement": "12.0.2"
   "18.0.0":
      "requirement": "18.0.3"
```
conanfile.py
```py
selected = self.conan_data.get("extra")[self.version]["requirement"]
```

or similar :)

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -373,7 +373,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_first_level = ["sources", "patches"]
+        allowed_first_level = ["sources", "patches", "extra"]
         allowed_sources = ["md5", "sha1", "sha256", "url"]
         allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5", "patch_type", "patch_source", "patch_description"]
         if conanfile.name in ["openssh", "gmp"]:
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -373,7 +373,6 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_first_level = ["sources", "patches", "extra"]
         allowed_sources = ["md5", "sha1", "sha256", "url"]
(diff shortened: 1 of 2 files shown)


uilianries:
@RubenRBS Would be possible to name the package reference instead? I looked the yaml first, without reading the description, with no context, I really didn't get the idea. My point is trying to make the yaml clear without needed to search in docs about the format or what is it about.

Could it be:

```yaml
sources:
  ...
  
patches:
  ...

requirements:
    "0.35.0":
      - "wasmtime/0.35.1"
```

So, we could:

```python
def requirements(self):
    for requirement in self.conan_data.get("requirements")[self.version]:
        self.requires(requirement)

# Now we move all dependencies from conanfile.pys in CCI to conandata.yml #joke
```

Of course, it could result in users abusing of it, instead of using conanfile.py. But we know it's for a specific case.

uilianries (approved):
In the very beginning, the conandata.yml had no format defined for non-regular cases, including packaging pre-built binaries. Users started to push different very different layouts for conandata.yml, making hard to track changes in the repository. To keep aligned most of layouts, we added a rule to check the conandata.yml format

In 2024, with Conan 2.x features and more than +1000 recipes, the ConanCenterIndex now follows a very common layout for each case. So when someone opens PR with a different layout, is because it's an exceptional case and deserves attention from reviewers to cover it.

PR #533, opened 2024-04-20
Avoid warning on applications

RubenRBS:
Could explicitly filter out by name, but I think this makes more sense

/cc @jcar87

rf https://github.com/conan-io/conan-center-index/pull/23582

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -882,6 +882,9 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
 
     @run_test("KB-H070", output)
     def test(out):
+        # Don't enforce it for applications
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -882,6 +882,9 @@ def test(out):
 
     @run_test("KB-H070", output)
     def test(out):
+        # Don't enforce it for applications
+        if getattr(conanfile, package_type, None) == "application":
```
RubenRBS, line 886:
```suggestion
        if getattr(conanfile, "package_type", None) == "application":
```

danimtb (approved):

uilianries (approved):
LGTM

PR #534, opened 2024-04-20
Add a linter exception for `gfortran`

valgur:
Add an exception for `gfortran` to the `DEFAULT PACKAGE LAYOUT` check.

Like other GCC packages, https://github.com/conan-io/conan-center-index/pull/23334 otherwise fails due to `libexec` being present in the package folder.

GCC uses `libexec` for internal tools and plugins that should not be exposed on PATH.

Perhaps `libexec` itself could be added to the `base_known_folders = ["lib", "bin", "include", "res", "licenses"]` list? Packages that rely on `libexec` usually do it for a good reason and reconfiguring the directory layouts can be difficult.


RubenRBS (approved):

danimtb (approved):

valgur:
Ping.

RubenRBS (approved):

PR #537, opened 2024-04-24
Relax conandata.yml hook for patches and make strict for url

RubenRBS:
* Allow customizing the fields for patches - additional can be okay if justified. If the minimal ones are not there, apply conandata patches will fail (and that should be enough of a check)
* For URLs, disallow checksums that are not `sha256` moving forward

### DIFF
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -378,7 +378,7 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_sources = ["sha256", "url"]
+        allowed_sources = ["sha256", "md5", "sha1", "url"]
 
         checksums = [ "sha256"]
         found_checksums = []
--- a/hooks/conan-center.py
+++ b/hooks/conan-center.py
@@ -378,18 +378,9 @@ def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_sources = ["md5", "sha1", "sha256", "url"]
-        allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5", "patch_type", "patch_source", "patch_description"]
+        allowed_sources = ["sha256", "url"]
 
-        extra_allowed_patches = {
-            "openssh": ["patch_os", "patch_os_version"],
-            "gmp": ["patch_os"]
-        }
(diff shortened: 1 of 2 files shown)


hooks/conan-center.py
```diff
@@ -378,18 +378,9 @@ def test(out):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_sources = ["md5", "sha1", "sha256", "url"]
-        allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5", "patch_type", "patch_source", "patch_description"]
+        allowed_sources = ["sha256", "url"]
```
Croydon, line 381: It would be nice if we would require `sha256` and `url`, but also allow `md5`, `sha1`

Some projects still only provide such a checksum and having the official checksum in there is desirable.
Furthermore, Conan will check all specified checksums, hence there is no security downgrade going on, but rather an improvement when specifying more than one.

hooks/conan-center.py
```diff
@@ -378,18 +378,9 @@ def test(out):
     def test(out):
         conandata_path = os.path.join(export_folder_path, "conandata.yml")
         version = conanfile.version
-        allowed_sources = ["md5", "sha1", "sha256", "url"]
-        allowed_patches = ["patch_file", "base_path", "url", "sha256", "sha1", "md5", "patch_type", "patch_source", "patch_description"]
```
Croydon, line 382: It would also be desirable if we could enforce `patch_type` and `patch_description` for all NEW patches / package versions, but no idea how this could be implemented reliable into the hooks.

hooks/conan-center.py
jcar87, line 381:
```suggestion
        allowed_sources = ["sha256", "md5", "sha1", "url"]
```

RubenRBS: You brought up a nice point, I had not considered the case where we would like to include published checksums as part of the conandata sources field alongside the `sha256` one - I've now implemented the option to optionally specify more than just that one, thanks a lot @Croydon!

hooks/conan-center.py
RubenRBS: While I personally would like everyone to have that, it's true that some recipes have not had them added yet, which would create problems for users that wanted to change unrelated things in such recipes with them being forced to add the extra fields when they had no initial intention/will to do so, so I think it's best if we don't enforce it for now :)

uilianries: We would need git diff, which is not covered by hooks, they just read the current recipe state :(

uilianries (approved):

danimtb (approved):

PR #538, opened 2024-04-25
[ci] Bump conan versions

danimtb:
Bumping the conan prev versions for both v1 and v2 since there where some releases behind.
The situation should not be an issue since both latest v1 and v2 plus both develop v1 and v2 versions were tested.


uilianries (approved):
LGTM

PR #543, opened 2024-07-05
fix: Add missing parameter

jakalx:
The call to `check_non_ascii` lacked the third parameter in case there is a `test_package/conanfile.py`.


uilianries (approved):
LGTM. Thank you for fixing it!

